PR #6, opened 2018-05-06
Remove entire codebase except README, LICENSE, dotfiles

mathcass:
@ericboucher, I decided to go and make this PR to see what you thought abot it. 

If we're going to copy the entire codebase into this, it might make sense to remove everything that exists here (and probably tag the commit prior to this one). 

The reason why I suggest this is because if we remove the old files and copy the new ones within a single commit, it could be weirder to look back through history. 

What do you think?


ericboucher:
I am fine with this to make it clearer. Did we decide on license? We use Apache2 for most of our OS projects I believe. An interesting ref for license choice is - https://exygy.com/which-license-should-i-use-mit-vs-apache-vs-gpl/

ericboucher:
Let me know when you have create a tag and I can approve this PR :)

mathcass:
Here it is, https://github.com/mapswipe/mapswipe/tree/pre-oss-app

That's a good read. We haven't decided on a license yet.

faevourite:
Hey folks. 👋 I'm trying to set up MapSwipe locally, so I could start hacking away on it and hopefully contribute some bug fixes. I'm going off commit caf4724d, since it's the last one that includes the actual React code. It looks like maybe you are about to do a bit of a reset here. :) Is there a different codebase I should start with instead? Thanks for any pointers!

mathcass:
@faevourite Yes! That's exactly what's going on here. Sorry for the interruption here. The open source version of the codebase had largely gone unmaintained compared with a private version so we decided to just go ahead and make the private version public. What we're preparing to do here is replace the entire contents of this repo with the private version. 

All hope isn't lost, though because I think we might be a week out from doing that. So, you should be able to come back to this then to take a look. 

As far as the code to pay attention to, pay attention to this repo and also this one, https://github.com/mapswipe/python-mapswipe-workers. Together, that's MapSwipe.

faevourite:
Awesome, thanks, @mathcass ! Good luck with the import. Is there a forum/chat where MapSwipe developers coordinate their efforts?

mathcass:
@faevourite Absolutely. We have a Slack channel. Could you email me at github@mathcass.com with an introduction?

PR #27, opened 2018-08-28
Update Firebase user events

ericboucher:
First attempt to update Firebase events, here are the changes:

- Add account_screen_seen when loading Login
- Add account_creation_error_username and account_creation_error_db for errors
- Remove misleading starting_tutorial
- Move starting_tutorial to the render section
- Add completed_tutorial

### DIFF
--- a/ios/Mapswipe/Info.plist
+++ b/ios/Mapswipe/Info.plist
@@ -15,11 +15,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.1.0</string>
+	<string>1.2.2</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>1.1.0</string>
+	<string>1.2.2</string>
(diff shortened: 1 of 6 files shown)


ios/Mapswipe/Info.plist
```diff
@@ -15,11 +15,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.1.0</string>
+	<string>1.1.1</string>
```
mathcass: Do you mind bumping this to 1.2.2. I bumped up to a minor version rather than a patch version with the OSS updates here: https://github.com/mapswipe/mapswipe/blob/baab45dd15534bbc292d6acc8728a09c46822bc7/android/app/build.gradle#L87-L88

We'll also need to increment the version number in build.gradle as well as increment the `versionCode`. I'm happy to increment the Android parts as needed and I'll probably need to do that before we merge anyway so we get the same version going.

mathcass: I've been meaning to do something like this https://medium.com/@andr3wjack/versioning-react-native-apps-407469707661 since I think it would simplify this a little bit

PR #41, opened 2019-03-16
RNSentry now brought in directly rather than via cocoapods due to com…

AlexCornforth:
The iOS app would not compile due to an issue with RNSentry. After doing some research it seemed this was an issue with linking RNSentry via cocoapods. This pull request adds and links the RNSentry framework to the project directly rather than via cocoapods and removes RNSentry from the iOS Podfile.


laurentS:
Hi @AlexCornforth sorry I've been way too slow at responding to your PR. We've had a proper look through the iOS build in the meantime to add build variants, and it's working decently now, so I'm going to close this one.
thanks for looking into it though!

PR #66, opened 2019-06-21
fix (more-options): Fix text on more options page for iOS

asaschachar:
The extra padding was cutting off the text inside the buttons. This may only be happening on iOS so a check was made specifically for that platform. It may make sense to try and move to a [React Native component library](https://blog.bitsrc.io/11-react-native-component-libraries-you-should-know-in-2018-71d2a8e33312) to get the benefits from a framework that has hopefully already solved these types of issues cross-platform visual issues.

Before:
Screen Shot 2019-06-21 at 11 21 27 AM

After:
Screen Shot 2019-06-21 at 11 21 47 AM

cc: @ericboucher

### DIFF
--- a/src/shared/views/MoreOptions.js
+++ b/src/shared/views/MoreOptions.js
@@ -9,6 +9,7 @@ import {
     ScrollView,
     StyleSheet,
     Image,
+    Platform,
 } from 'react-native';
 import Button from 'apsl-react-native-button';
 import * as Progress from 'react-native-progress';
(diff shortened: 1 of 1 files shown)


laurentS:
@asaschachar thanks for fixing this one, it's one of these annoying small things I can't fix, having no access to a macbook. Good point about using libs, though I've had a few days of hairpulling and headbanging during the last big upgrade of react-native, while trying to get all the dependencies to play together. So I'm a bit careful when adding a dependency for just a few lines of code. Will look into them though if we get more issues, or for refactoring of certain bits like buttons.

PR #67, opened 2019-06-21
feat (analytics): Added segment analytics events on mobile app

asaschachar:
This PR sends basic analytics events to [this segment project](https://app.segment.com/mapswipe/sources/mapswipe_mobile_app/overview) from the mobile application to track user behaviors like loading the application, toggling tiles, and completed tasks.

Events are being sent from the mobile app rather than through backend firebase functions because the mobile app does not communicate with the firebase backend for these actions.

[Steps taken to install segment analytics in react native](https://github.com/segmentio/analytics-react-native)

cc: @ericboucher

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -197,6 +197,12 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-splash-screen')
+    compile project(':react-native-sentry')
+    compile project(':react-native-linear-gradient')
+    compile project(':react-native-firebase')
+    compile project(':react-native-device-info')
+    compile project(':@segment_analytics-react-native')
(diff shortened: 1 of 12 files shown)


MegaArthur:
What about using some other analytics method than Firebase? FOSS and Google don't go together well

Hagellach37:
I'm going to close this PR as we have updated a bunch of dependencies since 2019. It is very likely that this has an impact on the changes proposed here. Also we need to have a discussion with the wider MapSwipe community about Google based analytics.

For now we will try to minimize such features where we can. Agreeing that Google and FOSS go not so well together and we've tracked some of the related things in issues already. (e.g. #120 )

PR #92, opened 2019-07-24
Version script

HHK1:


HHK1:
@laurentS I wrote the script we discussed this morning. It's updating both the xcodproj as well as the package.json. I'll let you update the android app correctly.

The version script must be execute from the root folder of the project, tell me if everything works fine for you.

For the tests I've tried several things without success. I think the next step is to open an issue on Travis, because I have no idea what's going on.

laurentS:
I just looked at the docs for travis, and they say in [here](https://docs.travis-ci.com/user/deployment#conditional-releases-with-on) that if `tags: true` is in the conditions, then the branch is ignored. But from their examples, it looks like we can do something like:
```
          on:
            # branch: master  # this is ignored because of the next line
            tags: true
            repo: mapswipe/mapswipe
            condition: $TRAVIS_BRANCH =~ ^master$
```
This would trigger the deployment only when branch is master and the commit is tagged. If it works, we could avoid the "beta" addition in the tag, which would keep it a little bit simpler.
The above would trigger the `release` deployment.
For all other branches, the condition to deploy the `beta` version would look like:
```
          on:
            # branch: master  # this is ignored because of the next line
            tags: true
            repo: mapswipe/mapswipe
            condition: $TRAVIS_BRANCH != master
```
What do you think?

HHK1:
@laurentS I thought about it, but according to the doc on $TRAVIS_BRANCH

>for builds triggered by a tag, this is the same as the name of the tag (TRAVIS_TAG).
See: https://docs.travis-ci.com/user/environment-variables/


So I think this would not work. Travis is considering a tag to be a branch, which is why I thought the tag itself should include some sort of differentiation between prod and beta

PR #138, opened 2019-10-15
Add some doc about git workflow

laurentS:
Here's a draft of the docs we discussed. @jhenshall @Hagellach37 comments welcome


jhenshall:
Actually, being pedantic and personal preference so feel free to ignore... can we have the prod and dev branches as a colour and both the features as grey - feel that would highlight the two versions of the app?

laurentS:
@jhenshall 
1. yes, the alphanumeric codes are called `git hashes` they are a unique reference to each commit. I wouldn't have added them, but the software I used just puts them there. If you click on the 'commits' tab at the top of this PR page, you'll see one linked to each commit. The ones in the doc are just fake ones.
2. fixed :)

Issue #46, opened 2019-05-01
Scalebar (for Built Area projects)


adrienandrem:
While browsing tiles, natural patterns can make you wonder what you are actually looking at.

A scalebar, displayed on the upper border for example, could be really useful.

laurentS:
Thanks for the feedback!

This has been requested by a number of people, and it's been included in the latest design updates. Hoping to get this out to users reasonably soon.

rivermont:
Copied from #111 


Since some projects have slightly different zoom levels, or because of offset and other imagery quirks it can be hard to tell the scale, I think it would be good to include a small ruler in one of the bottom corners of the viewport. Similar to what the iD editor has:

![image](https://user-images.githubusercontent.com/25139386/64914372-f969e180-d71d-11e9-9f82-ffb3d15c4846.png)

It would be overlayed on top of the tiles, but with transparency. Long-pressing the tile under it wouldn't show the scale on the enlarged image.

PR #142, opened 2019-10-17
#46 Add a scalebar to built area projects

laurentS:
Scalebar added on built area projects, here's what it looks like on my phone.
Screenshot_20191017-125010

It adjusts to the actual zoom level, and latitude, with preset scalebar sizes stolen from OSM
Screenshot_20191017-125000

### DIFF
--- a/src/shared/common/ScaleBar.js
+++ b/src/shared/common/ScaleBar.js
@@ -1,0 +1,114 @@
+// @flow
+import * as React from 'react';
+import {
+    ART,
+    View,
+} from 'react-native';
+import GLOBAL from '../Globals';
+
+type Props = {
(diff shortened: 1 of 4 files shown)


Hagellach37:
Hey,
I'm not sure if the upper image displays the scale bar in a correct size. I think 50 meters should cover much more. Heidelberg has a latitude of around 50 or so, so the entire tile width is around 100 meters. so 50 meters should cover half of a tile.

laurentS:
@Hagellach37 you were right, indeed. I got my units all mixed up. I've fixed the scale, so it now looks correct.
Any other issues ? If not, I'll merge into `dev`.

Screenshot_20191021-134644

PR #195, opened 2020-04-21
osm building layer added in Tile.js, and modified index.js instructi…

TahiraU:
@laurentS in this branch you will find following modifications:

- added osm building layer in Tile.js
- adjusted "How To Contribute" text field from "MAYBE" to "NOT COMPLETE" for completeness project type


src/shared/views/Mapper/Tile.js
```diff
@@ -143,8 +151,14 @@ export class _Tile extends React.Component<Props> {
         return { uri: tile.url };
     }
 
+    getOsmBuildings = () => {
```
laurentS, line 154: I would rename this `getOsmBuildingsUrl` or similar to clarify what the function returns.

src/shared/views/Mapper/index.js
```diff
@@ -237,9 +237,15 @@ class _Mapper extends React.Component<Props, State> {
                             YES
                         </Text>
                         , twice for&nbsp;
+
+                       {tutorialName == 'completeness_tutorial'?
```
laurentS, line 241: What happens if we have a custom tutorial name?

src/shared/views/Mapper/Tile.js
```diff
@@ -143,8 +151,14 @@ export class _Tile extends React.Component<Props> {
...
+        const { tile } = this.props;
+        return { uri: tile.urlB };
+    }
+
     zoomRender = () => {
         const imageSource = this.getImgSource();
+        const osmBuiling = this.getOsmBuildings();
```
laurentS, line 161: Typo: `osmBuiling` -> `osmBuildings` ? Maybe worth renaming it into something like `osmBuildingsImageSource`?

src/shared/views/Mapper/Tile.js
```diff
@@ -143,8 +151,14 @@ export class _Tile extends React.Component<Props> {
         return { uri: tile.url };
     }
 
+    getOsmBuildings = () => {
+        const { tile } = this.props;
+        return { uri: tile.urlB };
```
laurentS, line 156: I suspect `yarn flow` is going to complain about this new attribute on the `tile` object, you will probably need to update the type definitions.

PR #199, opened 2020-04-27
Prepare for release 2.0.3

laurentS:
This release does not change much from the user perspective:
- Scalebar should now be the correct size and be shown all the way to the end of a session
- "Built Area" projects can now use custom tutorials, per project if needed

The major part of the work went into upgrading React Native to version 0.61.5. This should be entirely transparent to users, but also moves the app's codebase on Android to use the new AndroidX system. This is a step that had to happen to ensure Google doesn't eventually stop accepting our updates.
Some minor warnings were fixed on iOS, in relation to the build version.

Overall, users should not even notice they're using a new version of the app.


ios/Podfile.lock
```diff
@@ -76,116 +86,156 @@ PODS:
...
-  - React (0.59.10):
-    - React/Core (= 0.59.10)
-  - react-native-netinfo (3.2.1):
-    - React
-  - react-native-splash-screen (3.2.0):
-    - React
-  - React/Core (0.59.10):
-    - yoga (= 0.59.10.React)
-  - React/CxxBridge (0.59.10):
+  - React (0.60.6):
```
Hagellach37, line 89: just for me to understand it:
you upgrade to react-native 0.61.5 or something like this, right? but why it is 0.60.6 here.

src/shared/views/ProjectView.js
```diff
@@ -331,6 +332,7 @@ class _ProjectHeader extends React.Component<HeaderProps, HeaderState> {
             projectType: project.projectType,
         });
         switch (project.projectType) {
+        case COMPLETENESS_PROJECT:
```
Hagellach37, line 335: hey,
just a quick one here: in case of COMPLETENESS_PROJECT nothing happens, right?

src/shared/views/ProjectView.js
```diff
@@ -510,7 +512,8 @@ class _ProjectHeader extends React.Component<HeaderProps, HeaderState> {
                             fb.analytics().logEvent('starting_tutorial', {
                                 projectType: project.projectType,
                             });
-                            if (project.projectType === LEGACY_TILES) {
+                            if ((project.projectType === LEGACY_TILES)
```
Hagellach37, line 515: this looks very similar to the code above (line 335/336): what's the difference?

src/shared/views/ProjectView.js
laurentS: This basically means "do the same thing as the following". See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch for reference. Here, COMPLETENESS_PROJECT and LEGACY_TILES will do the exact same thing.

src/shared/views/ProjectView.js
laurentS: Indeed. The code above starts the project mapping, while this one goes to the tutorial. I've rewritten this to use a `switch` instead in https://github.com/mapswipe/mapswipe/pull/199/commits/d18ace7876483515004b0df88e4f03435abe6495, so it looks a bit cleaner. We'll probably have to refactor that code after setting up the new Change Detection, as the logic is likely to change a fair bit.

ios/Podfile.lock
laurentS: Good catch. I think you saved the day here. It seems that I forgot to run the `pod install` command when upgrading to the latest RN, so it probably means that the test build was running on 0.60 on iOS, and 0.61 on android? Just upgraded pod files in https://github.com/mapswipe/mapswipe/commit/f0a797a4763a2b0827cc64dd1f47a1101ec62a87 In any case, we should probably do a bit more testing on iOS before releasing, just to make sure nothing is broken.

package.json
```diff
@@ -12,65 +12,74 @@
...
+    "@react-native-community/viewpager": "^3.3.0",
+    "@sentry/react-native": "^1.3.3",
     "apsl-react-native-button": "^3.1.1",
     "i18next": "^13.1.0",
     "lodash.get": "^4.4.2",
-    "react": "16.8.3",
+    "react": "16.9.0",
     "react-i18next": "^9.0.2",
-    "react-native": "0.59.10",
+    "react-native": "0.61.5",
```
Hagellach37, line 35: nice

src/shared/views/Mapper/Tile.js
```diff
@@ -151,15 +168,23 @@ export class _Tile extends React.Component<Props> {
                         borderColor: 'rgba(255,255,255,0.2)',
                     }}
                     source={imageSource}
-                />
+                >
+                    <ImageBackground
```
Hagellach37, line 172: It's a bit counter intuitive to have ImageBackgorund twice, since one of it is the overlay. But this is probably a react-native thing...

src/shared/views/Mapper/Tile.js
```diff
@@ -191,9 +233,9 @@ export class _Tile extends React.Component<Props> {
                     key={`touch-${taskId}`}
                     source={imageSource}
                 >
-                    <View style={[styles.tileOverlay, { backgroundColor: overlayColor }]} key={`view-${taskId}`}>
-                        {animatedRows}
-                    </View>
+
+                    {comp}
```
Hagellach37, line 237: I guess "comp" refers to completeness

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mapswipe",
   "version": "2.0.2",
-  "build": "0",
+  "build": "8",
```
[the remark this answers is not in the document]
laurentS: that's the one, yes!

Hagellach37:
regarding travis, it says that the last build failed. But this is the old one before you speeded it up? Because, actually you merged it in here, but still it timed out.

src/shared/views/Mapper/Tile.js
laurentS: yes, they need to be backgrounds, because we have the `Touchable` element in front of it, which responds to taps. At least it's my understanding!

laurentS:
> regarding travis, it says that the last build failed. But this is the old one before you speeded it up? Because, actually you merged it in here, but still it timed out.

I cancelled a lot of builds because so many were essentially building the same code. When this gets merged, it will rebuild again on CI, and once that one passes, I'll do the release commit, so we'll be sure it works before pushing out.

src/shared/views/Mapper/Tile.js
laurentS: It's "component", declared on line 207, and defined on 210 or 218. We could probably rename it and add typing annotations...

laurentS:
>     * with this release, you can have custom tutorials
Yes, they should work now, at least for project types 1 and 4.

>     * what's the status for completeness within this release? (I see that it is somehow there, but I'm not sure which parts of the functionality are in it already)
They should work, though I wouldn't release a project to production until we've done a bit more testing on the dev app.

Issue #167, opened 2019-12-13
Speed up CI build time


laurentS:
This is valid for both android and iOS, but more important for the latter. Many builds have failed on iOS due to travis timing out after an hour.

Options to speed things up:
- cache everything we can. `yarn`, but also `cocoapods` and maybe rubygems (used for fastlane among others). Travis [docs](https://docs.travis-ci.com/user/caching) explain how to do it.
- skip some parts of the build?

laurentS:
PR #201 brought the build+deploy time for iOS down to 37 minutes, which should be good enough for now, with enough margin from the 50 minute timeout that we don't need to worry about it for a while.
If really a problem, we might be able to shave off another minute or 2 by caching yarn dependencies a bit more aggressively.

PR #201, opened 2020-04-28
#167 Speed up travis build for iOS

laurentS:
The iOS build seems to have come down from 35-40 minutes to about 13, which should prevent timeouts when releasing to the app store.
Tricks:
- cache podfiles
- upgrade fastlane and cocoapods
- tell cocoapods to use their CDN < this seems to have brought the most savings, like a good 20 minutes


.travis.yml
```diff
@@ -9,6 +9,7 @@ git:
 cache:
   bundler: true
   cocoapods: true
+  podfile: ios/Podfile
```
Hagellach37, line 12: just asking here: if this is cached, what happens if we change something in that file? This will be recognized probably, right?

laurentS: I think the `pod install` command will simply notice that some files are outdated and overwrite them.
But if that doesn't happen, the build would probably fail, and you can go to https://travis-ci.org/github/mapswipe/mapswipe/caches and clear the cache for that branch, and rerun the CI job.

Issue #204, opened 2020-05-02
Reimplement Change Detection project based on new design


laurentS:
Rebuild the Change Detection project interface based on the user testing carried out by Chris and the design he proposed based on it.
This is related to #180, which has been a common complaint of users, namely the inability to "undo". The new design makes this much easier, and more similar to the UX users are used to.

laurentS:
Here are a few screenshots of the designs as they are right now. They might evolve, but this is what served as a base for the current work.
![image](https://user-images.githubusercontent.com/1745184/80955346-abaa0280-8dee-11ea-824f-726fc287658b.png)

![image](https://user-images.githubusercontent.com/1745184/80955377-bd8ba580-8dee-11ea-9834-5fcd3fa10304.png)

![image](https://user-images.githubusercontent.com/1745184/80955411-ce3c1b80-8dee-11ea-9399-157c84704deb.png)

![image](https://user-images.githubusercontent.com/1745184/80955447-dc8a3780-8dee-11ea-9b30-e755805c5acb.png)

PR #207, opened 2020-05-04
Rebuild Change Detection based on new design

laurentS:
This PR sets up the Change Detection (CD) project type based on the designs by Chris (see screenshots in #204).

So far the changes involve:
- rebuilding the CD project from scratch, on top of the native `FlatList` react-native component. There are several options to create a scrolling component, but this one seems by far the most resource efficient, especially around memory use.
- I have reused many existing bits of code I wrote before, which allowed going pretty fast on this.
- the core functionality should work: open a project, swipe and tap, complete the session and save results.
- tutorial mode is not supported yet, but will be added as part of this PR.
- I will push a few releases as part of this PR to test functionality across platforms
- help screens are not available yet.
- the goal is to then rebuild the original `Built Area` project type on top of that same component, which should help with some of the problems we've seen, like slow loading, high memory use and scrolling issues on "long" sessions.

This PR is made against the prettifier branch to make it easier to just focus on the actual functional changes, instead of all the formatting that happened between `dev` and `prettifier`.


Hagellach37:
Hey,
I'm just having a general comment here: I tried following the code and really like the fact that most things might be able to be reused by all project types. However, when it comes to this we might need to clarify some of the terms we are using. For instance:
* `SatImage` vs. `Tile` vs. `Task` or
* `Group` vs. `TaskList`
* not mentioning all the terms we use for the back end

I'm happy to support here, e.g. by creating an overview page and maybe also something more visual to get a more intuitive understanding of each. :)

laurentS:
@Hagellach37 this is a good point. I've already renamed the classes like 3 times over the weekend, to try to give them more meaningful names, but obviously it's not good enough yet! Diving back into my own code from not so long ago made me realise how little I had documented it. I think you're right that even just having a little sketch in the docs to explain how the components fit together would be great.

Hagellach37:
can we close this PR?

laurentS:
yes, it's definitely been merged through #221

PR #209, opened 2020-05-07
Add docs on how to use the test app

laurentS:
I wrote a short doc about how to install and run the test app, so that we can easily point volunteers to it. Can you have a quick read through and let me know if I've missed anything?

The bits I've removed are already on other pages in more details, so this was a bit redundant.


laurentS:
Closing this PR in favour of the wiki page https://github.com/mapswipe/mapswipe/wiki/How-to-test-the-development-version-of-MapSwipe which is easier to update

Issue #220, opened 2020-05-29
Improve memory usage (on Android?)


laurentS:
**Describe the bug**
Some users have reported memory usage in excess of 1GB on android. Completeness projects seem to be the worst offenders, though there is no profiling data available to confirm this yet.

It should be possible to reduce this, which should help a lot with user engagement on low end devices.

**App version**
2.0.3(9) and the few versions just before.

**Smartphone (please complete the following information):**
Android, not clear on iOS.

**Additional context**
As of React-Native 0.60+, a new JS engine is shipped with RN (Hermes) to help improve memory consumption of the old JSC that was (is still) used on android. I would be worth trying it out properly to see if there's any improvement.
[This post](https://medium.com/walmartlabs/react-native-memory-profiling-jsc-vs-v8-vs-hermes-1626949a653b) also suggests another option (react-native-v8) which might make it even better.

Issue #219, opened 2020-05-28
completeness project is sticky after swiping 3 screens


Hagellach37:
**Describe the bug**
The app is just super slow after a few screens of the new completeness project type in the dev version of the app.

**App version**
iOS, 2.0.3(9)-beta

**To Reproduce**
Please provide steps to reproduce the problem:
1. Go to project Nuernberg Completeness test - (1)
2. Click on "Map Now"
3. Scroll left to the next screens and see the app becoming really sticky after 3 screens or so untill the end of the group
4. at some point you even reach a point where imagery is not loaded.

**Smartphone (please complete the following information):**
 - Samsung Galaxy Tab A (2018, 10.5)

**Additional context**
The app seems to have a fairly high usage of RAM, on average around 250MB, max to 1.2GB on the tablet.

Things we could try:
* do not use opacity in react-native for image background but in mapbox instead
* reduce group size during project creation
* take another look at the flatlist component

laurentS:
Using the performance monitor of react-native (in developer tools), I can see that the FPS stays above 50 for built-area projects pretty much all the time, even when scrolling too fast to do anything.
In completeness projects, the FPS drops to below 20, I think this is what you are seeing, but maybe with worse numbers.

[this page](https://reactnative.dev/docs/0.61/optimizing-flatlist-configuration) has a few tips about improving the rendering of flatlist. Playing with `windowSize` and `maxToRenderPerBatch` is helping a bit on the FPS counter.

Hagellach37:
After testing on my tablet I can confirm that the `windowSize` parameter has a big effect on the performance.

windowSize={5} --> ~50fps, nice and fast
windowSize={15}  --> dropping to 10% of that (5fps), slow and laggy

laurentS:
This component could be helpful as well: RecyclerView https://github.com/Flipkart/recyclerlistview

or another option could be to test different JS engines on android: https://medium.com/walmartlabs/react-native-memory-profiling-jsc-vs-v8-vs-hermes-1626949a653b

laurentS:
Fixed in 2.0.4

Issue #217, opened 2020-05-27
username is missing in "public" part of user profiles


laurentS:
## This has no impact on the app itself!!

**Describe the bug**
The username is not copied to the profile part of a user account data on firebase when registering a new account. When the profile is initialised in `handleSignup`, a number of properties are set, but the `username` is missing. It is correctly set in the conversion from v1 to v2 though.

The missing usernames are not found in the leaderboard.

**App version**
This seems to date back to a commit in Nov 2018.

laurentS:
Fixed in 2.0.4

Issue #197, opened 2020-04-22
Show app version number in the "More" screen


laurentS:
This would help testers and general bug reporting.
Need to find out how to collect the version number (and tag) from git during the app build process.
Thanks Chris for the idea!

laurentS:
Implemented in 2.0.4

PR #226, opened 2020-06-05
Prepare for 2.0.4

laurentS:
Merge all the changes since 2.0.3. Lots of changes:

- Built area and Change Detection #204 projects are rebuilt using a FlatList component which gives:
  + smoother scrolling #219 #172
  + faster initial imagery loading #179
  + lower memory usage #220
- swipe down action is now much easier to access #127
- push notifications work on android and iOS #177
- usernames are saved to user profiles so they will show up in the leaderboard #217
- "continue mapping" button is enabled again #53
- debug info shown in the "more" screen #197


ios/mapswipeUITests/Info.plist
```diff
@@ -17,6 +17,6 @@
 	<key>CFBundleShortVersionString</key>
 	<string>2.0.3</string>
```
Hagellach37, line 18: do we need to change something here?

package.json
```diff
@@ -1,7 +1,7 @@
 {
```
Hagellach37, line 1: and here? should it be `2.0.4`?

PR #237, opened 2020-06-21
Remove timestamp value

### LINKED ISSUE #227: Remove "timestamp" attribute from result objects
We store a `timestamp` on the result object which seems to be the exact same as the `endTime` value, what is the point of storing both?

laurentS:
This PR removes a few `timestamp` variables that were either:
- misnamed, and I renamed them to `startTime` for more clarity
- useless, the value was passed in the action, but never actually stored, not even locally
- not wanted in firebase, this is what #227 is really about
Fixes #227

### DIFF
--- a/src/shared/actions/index.js
+++ b/src/shared/actions/index.js
@@ -199,15 +199,13 @@ export function commitGroup(groupInfo: GroupInfo): ThunkAction {
         const firebase = getFirebase();
         const userId = firebase.auth().currentUser.uid;
         // get a single timestamp upon completion of the group
-        const timestamp = GLOBAL.DB.getTimestamp();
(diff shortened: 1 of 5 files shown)


laurentS:
> Before wen merge this into dev, we need to make sure that the back end can handle the renamed attributes.

The commits here don't change anything that the app sends to the backend. It only changes their names internally, and removes the `timestamp` key that is sent along with `startTime` and `endTime` in the results. So there are no new attributes. However if the backend relies on `timestamp` then you should change that to use `endTime` instead, as their values are the same.

Hagellach37:
okay, the back end uses all attributes, but this is easy to adjust since `endTime` = `timestamp`.

PR #238, opened 2020-06-21
#236 Update iOS dev env setup docs

### LINKED ISSUE #236: Improve dev setup instructions for iOS
**Describe the bug**
The [instructions](https://github.com/mapswipe/mapswipe/blob/master/docs/develop-ios.md) for building and running the dev app on iOS are not good enough, as they are confusing to newcomers.
We should clarify what is actually needed to get a first build running, without mixing in parts that relate to release builds.

There might also be a way to simplify the list of required files for a dev build, so that production copies of files are not stopping the build, etc...

laurentS:
Fixes #236 
This updates the docs for iOS dev env setup, hopefully more in line with what is actually needed. I've removed the parts related to deployment as they are not needed to get started.

@oiva happy to get your feedback on this :)

### DIFF
--- a/docs/develop-ios.md
+++ b/docs/develop-ios.md
@@ -26,27 +26,32 @@ You should request access to the `dev-mapswipe` firebase instance, on which ther
 
 Next, download the Firebase Google Services files from Firebase > Settings > Add App > Download files.
 
-Copy the iOS file to `ios/cfg/GoogleService-Info.plist`.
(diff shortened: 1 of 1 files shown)


oiva:
Nice, it's a lot clearer now.

> Contact one of the project admins for this (see https://www.mapswipe.org/ for contacts).

Could this be replace with an email address? The MapSwipe website doesn't have a clear contact page except the email address found in the footer.

While setting up the development environment I also ran into a [code signing error](https://developer.apple.com/library/archive/qa/qa1940/_index.html) which required running `xattr -cr .` in the iOS path. I'm not sure if this is a universal problem or if I just messed something up.

PR #239, opened 2020-06-22
Live updates for project list

### LINKED ISSUE #93: Users get kicked out of a group if another user completes a group while they're mapping
This is essentially the same bug as #62.

#62 was caused by changes in the `group` key during mapping, whereas this issue seems related to updates to the `project` key, but both have the same effect.

### LINKED ISSUE #119: Subscribe to updates from firebase on project list
Following up on #93 , we should implement a nicer solution using firebase's subscription mechanism (via `react-redux-firebase`) so that list of projects, and progress are updated live, but without causing the app to reset ongoing mapping.

Possible improved solutions include (from https://github.com/mapswipe/mapswipe/issues/93#issuecomment-515926109):
- adding a listener for the `onBlur` and `onFocus` events of the navigator, to watch/unwatch changes, while mapping (https://reactnavigation.org/docs/en/navigation-prop.html#addlistener-subscribe-to-updates-to-navigation-lifecycle)
- but there must be a cleaner way to do this, so that changes in the project data don't trigger rerendering/rerouting when only the `progress` and similar attributes are updated.

laurentS:
This PR brings back live updates for the project list. We used to see progress and number of mappers increase while on the list of project, but had to disable this to fix #93. Not anymore!

Fixes #119

### DIFF
--- a/src/shared/views/RecommendedCards.js
+++ b/src/shared/views/RecommendedCards.js
@@ -6,6 +6,7 @@ import { firebaseConnect, isLoaded, isEmpty } from 'react-redux-firebase';
 import { ScrollView, StyleSheet, Text } from 'react-native';
 import Button from 'apsl-react-native-button';
 import Modal from 'react-native-modalbox';
+import type { NavigationEventSubscription } from 'react-navigation';
 import ProjectCard from './ProjectCard';
 import LoadingIcon from './LoadingIcon';
 import type { NavigationProp, ProjectType } from '../flow-types';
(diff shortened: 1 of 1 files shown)


src/shared/views/RecommendedCards.js
```diff
@@ -77,13 +78,92 @@ type OrderedProject = {
...
+        this.willFocusAnnouncementSubscription.remove();
+        this.willBlurAnnouncementSubscription.remove();
+        this.willFocusProjectSubscription.remove();
+        this.willBlurProjectSubscription.remove();
+    }
+
+    subscribeToAnnouncements = () => {
+        const { type, path, storeAs, ...options } = announcementQuery;
+        const { firebase, navigation } = this.props;
+        this.willFocusAnnouncementSubscription = navigation.addListener(
```
Hagellach37, line 136: What are `willFocus` and `willBlur` refering to? 

Is it like this?
*  users opens the project view, this "connects" to firebase projects and the screen is updated live
* when user selects a project and starts mapping, this triggers `willBlur` and the projects are disconnected from firebase

Hagellach37:
once you select a project you will see the `projectView.js` screen. I'm not sure how the changes here also apply to this screen. The projectView screen also holds the information on users and progress per project.

Is this already covered then?

src/shared/views/RecommendedCards.js
```diff
@@ -77,13 +78,92 @@ type OrderedProject = {
...
+        );
+        this.willBlurProjectSubscription = navigation.addListener(
+            'willBlur',
+            () => {
+                firebase.unWatchEvent(type, path, storeAs, options);
+            },
+        );
+    };
+
     closeModal3 = () => {
```
Hagellach37, line 167: which element is `Modal3`. Is there a better name?

laurentS: I am not sure, but as I look around the code, it's pretty clear that I need to spend some time renaming variable, functions and classes, the current names are rather confusing...

src/shared/views/RecommendedCards.js
laurentS: `willFocus` is triggered when the project list is about to be shown, and we then connect to firebase to get live updates on each change to the query we've sent. This continues until we `unWatch` it, which happens when the project list screen loses foces, in `willBlur`. Then we stop listening to firebase, so whatever happens on the db will not impact us.

laurentS:
> What would happen in the app if a project is set from `active` to `finished` in firebase?
> 
>     * if a) the user sees the project overview screen

then the project in question will disappear from the list

>     * if b) the user is mapping a group of a project which is set to `finished`

nothing while the user is mapping, as the app has stopped listening for updates. It will only "hear about it" once the user goes back to the project list.

laurentS:
> once you select a project you will see the `projectView.js` screen. I'm not sure how the changes here also apply to this screen. The projectView screen also holds the information on users and progress per project.

Right now, it will be like before, no updates happen while we're looking at a single project. The update will only come in when we navigate back to the main list. We could add a listener for the project in question, but I'm not sure it's worth the added complexity in the code. What do you think?

Hagellach37:
>  but I'm not sure it's worth the added complexity in the code. What do you think?

I think it's maybe not worth it. You will not see the project page if you select "continue mapping". Only if you select "stop mapping" at the end of the group you will be directed back to the project details page.

not super urgent, and much more important and cool to update projects on the main page.

PR #247, opened 2020-06-25
#246 Fix broken custom TMS imagery url calculation

### LINKED ISSUE #246: Some custom TMS layers don't load
**Describe the bug**
A TMS layer with a url template that includes `{-y}` won't load, because the code to calculate the correct url is buggy.

**App version**
2.0.4(0)

**To Reproduce**
1. Create a project with a custom TMS imagery layer that include `{-y}` in it.
2. Open the project in the app
3. No imagery is shown

laurentS:
Fixes #246 

@Hagellach37 I've figured this out by trial and error, but please check that it actually makes sense with the way the backend creates projects, and overall TMS definitions.

### DIFF
--- a/src/shared/common/tile_functions.js
+++ b/src/shared/common/tile_functions.js
@@ -64,7 +64,7 @@ export function getTileUrlFromCoordsAndTileserver(
         // https://www.maptiler.com/google-maps-coordinates-tile-bounds-projection/
         const newY = 2 ** zoom - y - 1;
         url = formatXYZoomKey(urlTemplate, x, newY, zoom, apiKey);
-    } else if (tileServerName === 'custom' && urlTemplate.includes('{-y}')) {
+    } else if (urlTemplate.includes('{-y}')) {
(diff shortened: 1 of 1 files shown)


src/shared/common/tile_functions.js
```diff
@@ -64,11 +64,11 @@ export function getTileUrlFromCoordsAndTileserver(
         // https://www.maptiler.com/google-maps-coordinates-tile-bounds-projection/
         const newY = 2 ** zoom - y - 1;
         url = formatXYZoomKey(urlTemplate, x, newY, zoom, apiKey);
-    } else if (tileServerName.includes('{-y}')) {
+    } else if (tileServerName === 'custom' && urlTemplate.includes('{-y}')) {
```
Hagellach37, line 67: I think we could just use `else if (urlTemplate.includes`... here. We don't have to check for the tileServerName attribute.

laurentS: ok, updated!

Issue #196, opened 2020-04-22
Capitalise app name on iOS


jhenshall:
**Describe the bug**
The app name is all lowercase on iOS home screen rather than capitalised as MapSwipe. This has been fixed in the dev version of app but not prod.

**Screenshots**
![IMG_4729](https://user-images.githubusercontent.com/6359748/79974495-56d2c780-8491-11ea-8865-ddec63a55325.jpeg)

laurentS:
I think this is done by now. @jhenshall please reopen if I'm mistaken.

jhenshall:
Yup, confirmation all done.

PR #250, opened 2020-06-27
Add BUNDLE_DISPLAY_NAME for production release

oiva:
Change the app's name to "MapSwipe" #196

ios/mapswipe/Info.plist seems to use $(BUNDLE_DISPLAY_NAME) variable as
_CFBundleDisplayName_ which is the name shown on users' home screen as the
app name.

The beta version bundle display name is defined in project.pbxproj
(BUNDLE_DISPLAY_NAME = "MapSwipe-beta";) but the release version did not
have BUNDLE_DISPLAY_NAME defined.

Not 100% sure if this will change the app name on iOS as I don't think I can produce production builds to test it. I did test adding the same variable for debug build and it changed the debug app's name in simulator.

### DIFF
--- a/ios/mapswipe.xcodeproj/project.pbxproj
+++ b/ios/mapswipe.xcodeproj/project.pbxproj
@@ -852,6 +852,7 @@
 			baseConfigurationReference = 1E0592B88B5096FD530933DC /* Pods-mapswipe.release.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				BUNDLE_DISPLAY_NAME = "MapSwipe";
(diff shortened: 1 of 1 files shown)


laurentS:
@oiva this is great! we've been meaning to find the solution to this for a while. Having to clear idea how to properly test this, I'm tempted to just merge your PR, and we should be able to see if:
1. the CI build passes
2. once we do the next release, the app first gets into limited availability, so @jhenshall should be able to check that it works for him, for instance.
I'm not sure how else we can test this with 100% certainty

jhenshall:
This is awesome, thank you @oiva ! Will leave the merge with @laurentS but when done, we can install the prod version of the app through TestFlight to test before pushing the actual release.

PR #255, opened 2020-07-08
Update LoadMore.js

kmpoppe:
Beautify the Sync Alert message.

I would have added the text to the en.json for translation but I wasn't able to find a file with the currently used 70 strings.

PR #256, opened 2020-07-09
Beautify sync alert

kmpoppe:
Rebased change from #255 to the now quite recent dev-branch.


kmpoppe:
@laurentS As I can't find other modes of contact for you: : could you supply a built Dev version of the app that I can test?

laurentS:
I've just added my email to my github profile :) but yes, I am working on a new dev version that should be out before the end of today, so that everyone can see their translations in context.

PR #257, opened 2020-07-09
Validate localization file JSON format

laurentS:
This PR adds a simple check in travis to validate the `locales/en.json` file format. This is to help avoid problems with transifex importing the file from github and silently complaining but not actively telling us there is a problem.

This should allow any developer to run `yarn lintJson` to check things locally, and it also runs on every build (the android one only, but it's enough to catch errors).


package.json
```diff
@@ -10,6 +10,7 @@
     "detoxTestIOS": "detox build -c ios.sim.debug && detox test -c ios.sim.debug",
     "reduxDebugger": "remotedev-debugger --hostname localhost --port 5678 --injectserver",
     "lint": "node_modules/eslint_d/bin/eslint.js src __tests__ --cache",
+    "lintJson": "node_modules/.bin/jsonlint -q locales/en.json",
```
Hagellach37, line 13: Should this check only `locales/en.json` or also all the other files that will eventually be in that folder such as `fr.josn`...?

Other than this I agree that it is a good idea to add test in travis.

laurentS: I only check `en.json` because it's the only one that should be manually modified by devs. The other languages are sent over by transifex, so I expect them to be valid json. That said, it could be worth checking them as well, just in case someone adds a manual fix. I'll open a separate PR.

PR #265, opened 2020-07-13
#264 Add support for private projects

### LINKED ISSUE #264: Add support for private projects
This is the frontend counterpart to https://github.com/mapswipe/python-mapswipe-workers/issues/392

The app needs to show private projects, initially instead of public ones. In the future, we might add an option for project managers or team managers to allow showing both private and public projects, but this would require highlighting them differently so it is clear to users what is what.

This ticket only requires checking if the user is part of a team, and adjust the query to the server if so.

laurentS:
This PR fixes #264 by adding basic support for private projects.

If the user is part of a team, the app simply pulls private projects assigned to their team. The user can confirm if they're part of a team by checking their profile page, which shows the team name if relevant, nothing otherwise.

### DIFF
--- a/src/shared/flow-types.js
+++ b/src/shared/flow-types.js
@@ -54,7 +54,7 @@ export type SingleImageryProjectType = {
     // FIXME: we should use constants here, somehow flow is not happy with them
     projectType: 1 | 2 | 4,
     progress: number,
-    state: number,
+    status: string,
     tileServer: TileServerType,
     tileServerB: ?TileServerType,
     zoomLevel: number,
@@ -72,7 +72,7 @@ export type ChangeDetectionProjectType = {
     projectId: string,
     projectType: 3,
     progress: number,
-    state: number,
+    status: string,
     tileServerA: TileServerType,
     tileServerB: TileServerType,
 };
--- a/src/shared/views/MoreOptions.js
+++ b/src/shared/views/MoreOptions.js
@@ -371,9 +371,7 @@ const mapStateToProps = (state, ownProps) => ({
     profile: state.firebase.profile,
     progress: state.ui.user.progress,
     teamId: state.ui.user.teamId,
-    teamName: state.firebase.data.teamDetails
-        ? state.firebase.data.teamDetails.teamName
-        : undefined,
+    teamName: state.firebase.data.teamName,
 });
 
 const enhance = compose(
--- a/src/shared/views/MoreOptions.js
+++ b/src/shared/views/MoreOptions.js
@@ -135,6 +135,8 @@ type MOProps = {
     profile: Object,
     progress: number,
     t: TranslationFunction,
+    teamId: ?string,
+    teamName: ?string,
 };
 
 // eslint-disable-next-line react/prefer-stateless-function
@@ -225,6 +227,8 @@ class _MoreOptions extends React.Component<MOProps> {
             profile,
             progress,
             t,
+            teamId,
+            teamName,
         } = this.props;
         const levelObject = Levels[level];
         const contributions =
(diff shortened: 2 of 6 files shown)


src/shared/views/RecommendedCards.js
```diff
@@ -149,7 +151,19 @@ class _RecommendedCards extends React.Component<Props> {
 
     subscribeToProjects = () => {
         const { type, path, storeAs, ...options } = projectsQuery;
-        const { firebase, navigation } = this.props;
+        const { firebase, navigation, teamId } = this.props;
```
Hagellach37, line 154: If I get this right this overwrite the existing query to get the projects from firebase, but I have problems to find the file where we would query the projects normally. Can you point me to it?

I think it's fine, but just want to understand the initial query as well.

src/shared/flow-types.js
```diff
@@ -54,7 +54,7 @@ export type SingleImageryProjectType = {
     // FIXME: we should use constants here, somehow flow is not happy with them
     projectType: 1 | 2 | 4,
     progress: number,
-    state: number,
+    status: string,
```
Hagellach37, line 57: interesting. why do we have to change it now? didn't we deprecate the state attribute already when switching to v2?

src/shared/views/RecommendedCards.js
```diff
@@ -81,6 +81,7 @@ type Props = {
     firebase: Object,
     navigation: NavigationProp,
     projects: Array<OrderedProject>,
+    teamId: ?string,
```
Hagellach37, line 84: what does the `?` stand for?

src/shared/views/MoreOptions.js
```diff
@@ -354,6 +370,10 @@ const mapStateToProps = (state, ownProps) => ({
     navigation: ownProps.navigation,
     profile: state.firebase.profile,
     progress: state.ui.user.progress,
+    teamId: state.ui.user.teamId,
+    teamName: state.firebase.data.teamDetails
+        ? state.firebase.data.teamDetails.teamName
```
Hagellach37, line 375: where does `teamDetails` come from?

src/shared/views/RecommendedCards.js
```diff
@@ -164,6 +178,15 @@ class _RecommendedCards extends React.Component<Props> {
         );
     };
 
+    getTeamName = () => {
+        // request the team display name from firebase
+        // the result is then available under state.firebase.data.teamDetails
+        const { firebase, teamId } = this.props;
+        if (teamId) {
+            firebase.watchEvent('once', `v2/teams/${teamId}`, 'teamDetails');
```
Hagellach37, line 186: there is no `teamDetails` endpoint in firebase. I'm not understanding what you want to do here. I might be a bit confused and you might set the `state.firebase.data.teamDetails` endpoint somewhere else, but where?

src/shared/flow-types.js
laurentS: We did, but somehow this state attribute wasn't seen by flow until I explicitly checked for `status`, and then it complained, which prompted me to change it. Basically, this line was never really used, it seems...

src/shared/views/RecommendedCards.js
laurentS: It's `flow`'s syntax. Basically means it's optional, ie, can be string or undefined.

src/shared/views/RecommendedCards.js
laurentS: Yes, this is the part that was modified to allow for "live" update of projects, so it's done in a slightly different way. On line 156 below, we call `firebase.watchEvent` which sets up a listener for changes from firebase. The details of the query are on line 91 above, in the same file. The `willFocus` and `willBlur` events are fired by the navigation when the user enters/leaves the screen with the list of projects.

src/shared/views/RecommendedCards.js
laurentS: `teamDetails` is where I tell the firebase client to store the data it downloads from the backend. This is only a local value, inside the app's data storage. The line above says "fetch `v2/teams/${teamId}` only once and place the result in the local firebase tree under `..../teamDetails/`. Then it's available from anywhere in the app under that same location (which is your other question above).

src/shared/views/MoreOptions.js
laurentS: See my other comment a bit further down on this page.

Issue #240, opened 2020-06-22
Tiles incorrectly displayed on very long screens


laurentS:
**Describe the bug**
A user reported a bug with the display of tiles in built area projects. See screenshot below. The tile at top right somehow moved to a 4th row at the bottom.

**App version**
`2.0.4(0)` on android

**To Reproduce**
1. Open any built area project
2. see the problem

**Screenshots**
![image](https://user-images.githubusercontent.com/1745184/85285199-fefc1100-b47f-11ea-8141-dd45735bab70.png)

**Smartphone (please complete the following information):**
 - Device: [OnePlus7T
 - OS: Android 10

**Additional context**
The screen size is 2400x1080.
Debug info give the following screen metrics in the app:
```
TILE_VIEW_HEIGHT: 895.89
SCREEN_WIDTH: 454.73
SCREEN_HEIGHT: 965.89
TILE_SIZE: 227.36
```
This seems to be caused by an incorrect flow configuration for the tiles.

kmpoppe:
Merge with #241, so this Issue is done and could be closed, right?

laurentS:
We try to close issues only once the corresponding features/bug has been merged into production (ie `master`)

laurentS:
Fixed with release `2.0.5(0)`.

PR #369, opened 2020-08-11
Prepare release 2.0.5(0)

laurentS:
This PR merges changes into `master` for the next production release.
Changes include:
- Multilanguage support (all text except tutorial prompts). Languages with partial translations around 50% or more are included, hopefully this might prompt more users to help
- Add support for private projects (#264)
- Fix bug where very elongated screens display tiles incorrectly (#240)
- Prevent duplicate data from being sent back to server upon group completion (#227)
- Some minor technical upgrades, like dependencies and certificates


src/shared/actions/index.js
```diff
@@ -193,15 +199,13 @@ export function commitGroup(groupInfo: GroupInfo): ThunkAction {
...
-        const timestamp = GLOBAL.DB.getTimestamp();
-        const endTime = timestamp;
+        const endTime = GLOBAL.DB.getTimestamp();
         const { groupId, projectId, results } = groupInfo;
         dispatch(startSendingResults(projectId, groupId));
         const { startTime, ...rest } = results[projectId][groupId];
         const objToUpload = {
             startTime,
             endTime,
-            timestamp,
```
Hagellach37, line 204: okay so from this release on there will be no `timestamp` attribute in the results anymore. I just double checked that this is supported by the back end. It's safe to remove here, the firebase function do not expect a timestamp attribute anymore.

src/shared/constants.js
```diff
@@ -23,3 +23,20 @@ export const COLOR_WHITE = 'white';
 export const COLOR_DARK_GRAY = '#212121';
 export const COLOR_LIGHT_GRAY = '#eef2fb';
 export const COLOR_TRANSPARENT_LIGHT_GRAY = 'rgba(238, 242, 251, 0)';
+
+// languages
+export const supportedLanguages = [
```
Hagellach37, line 28: do you add languages manually here? Or is this also handled by transifex?

src/shared/i18n.js
```diff
@@ -1,20 +1,37 @@
 // @flow
 import i18n from 'i18next';
```
Hagellach37, line 2: same question: this is probably handled by transifex as well?

src/shared/reducers/ui.js
```diff
@@ -14,6 +15,7 @@ import type { UIState } from '../flow-types';
 const defaultUserState = {
     hasSeenHelpBoxType1: false,
     kmTillNextLevel: 0,
+    languageCode: 'xx',
```
Hagellach37, line 18: should the default language code be `xx` or `en`. I'm not sure if it is even used, but xx could be confusing.

src/shared/reducers/ui.js
```diff
@@ -14,6 +15,7 @@ import type { UIState } from '../flow-types';
 const defaultUserState = {
     hasSeenHelpBoxType1: false,
```
Hagellach37, line 16: what is this referring to? the initial language selection screen?

src/shared/views/ChangeDetection/InstructionsScreen.js
```diff
@@ -143,84 +145,99 @@ export default class CDInstructionsScreen extends React.Component<Props> {
...
-                    <Text style={styles.header}>Your task:</Text>
+                    <Text style={styles.header}>{t('your task')}</Text>
                     <Text style={styles.tutParagraph}>
-                        You&apos;re looking for{' '}
-                        <Text style={{ fontWeight: 'bold' }}>
-                            changes in buildings
-                        </Text>
-                        . This acts as a clear indicator for a change in
-                        population size.
+                        <Trans i18nKey="CDInstructionsScreen:lookingFor">
```
Hagellach37, line 155: is this the long version of writing `{t('CDInstructionsScreen:lookingFor')}`. Above you always wrapped the text blocks into the t function. Here it looks differently.

src/shared/views/ChangeDetection/InstructionsScreen.js
```diff
@@ -143,84 +145,99 @@ export default class CDInstructionsScreen extends React.Component<Props> {
...
+                        </Trans>
                     </Text>
                     <Text style={styles.header}>&nbsp;</Text>
                 </ScrollView>
             </View>
         );
     }
 }
+
+export default withTranslation('CDInstructionsScreen')(CDInstructionsScreen);
```
Hagellach37, line 243: I'm a bit confused by this file and don't know how it is used. It seems to be related to instructions for change detection, but also somehow to the tutorial. 
I thought that we still load information on the tutorial from firebase. Is this the first step towards moving the tutorial into the app and make it translatable?

src/shared/views/LoadMore.js
```diff
@@ -117,29 +117,27 @@ class _LoadMoreCard extends React.Component<Props> {
...
+                    {tutorial ? t('completedTutorial') : t('finishedGroup')}
                 </Text>
 
                 <Button
                     style={styles.moreButton}
                     onPress={this.onComplete}
                     textStyle={{ fontSize: 18, color: COLOR_WHITE }}
                 >
-                    {tutorial ? "Let's go!" : 'Complete Session'}
+                    {tutorial ? t('letsGo') : t('completeSession')}
```
Hagellach37, line 132: this is the way of writing?
```
if tutorial then "let's go"
else "complete"
```

src/shared/views/Login.js
```diff
@@ -474,24 +482,24 @@ class _Login extends React.Component<Props, State> {
...
-                            onPress={() => {
-                                navigation.push('WebviewWindow', {
-                                    uri: 'https://mapswipe.org/privacy',
-                                });
-                            }}
-                        >
-                            Privacy Notice
-                        </Text>
+                        <Trans i18nKey="signup:IagreeToPrivacyNotice">
+                            I agree to the bal
```
Hagellach37, line 486: `bal` ?

src/shared/views/Mapper/index.js
```diff
@@ -217,89 +220,84 @@ class _Mapper extends React.Component<Props, State> {
...
-                        once for&nbsp;
-                        <Text style={{ color: 'rgb(36, 219, 26)' }}>YES</Text>,
-                        twice for&nbsp;
-                        {comp}, and three times for&nbsp;
-                        <Text style={{ color: 'rgb(230, 28, 28)' }}>
-                            BAD IMAGERY (such as clouds)
-                        </Text>
-                        .
+                        <Trans
+                            i18nKey="Tutorial:instructions3"
```
Hagellach37, line 248: again here it looks like part of the tutorial have been put into the app.

src/shared/views/RecommendedCards.js
```diff
@@ -77,13 +78,119 @@ type OrderedProject = {
...
     navigation: NavigationProp,
     projects: Array<OrderedProject>,
+    teamId: ?string,
+};
+
+// request only active projects from firebase (status === 'active')
+// firebase doesn't allow multiple query params, so for project types we filter in render()
+// but here we can still limit to 20 projects maximum
+// `path` defines where the resulting data is copied in the redux store
+// (state.firebase.ordered.projects in this case, because we've asked for `orderByChild`)
```
Hagellach37, line 91: as a general comment we could add here that this query needs to be allowed by the firebase db rules set in the back end

src/shared/constants.js
laurentS: We have to "activate" them manually in the app. This way transifex can send translations for a language that only has very limited content, and we can delay turning it on in the app. I don't think there's an easy way to automate this, unless we look at the files in the locales folder, and check how many lines they contain and all. I think it's not worth the effort, at least not now.

src/shared/i18n.js
laurentS: no, same here, we need to declare the languages we want in the app in both places (here and your other comment).

src/shared/reducers/ui.js
laurentS: I used `xx` here to help detect that no language has been selected by the user. I guess I could have set it to `undefined` instead. Will fix in next release, just in case I mess something up. https://github.com/mapswipe/mapswipe/issues/370

src/shared/reducers/ui.js
laurentS: it's a flag set to true the first time you see the help box for project type 1 (`built_area`). the naming is not the best, indeed. I've added a short comment in the code.

src/shared/views/ChangeDetection/InstructionsScreen.js
laurentS: Almost. This `Trans` component is used when there are nested components, which you could not include in the `{t('...')}` function. See https://react.i18next.com/latest/trans-component for details

src/shared/views/ChangeDetection/InstructionsScreen.js
laurentS: This is the static text that is in the help screen you get to when you tap the `i` icon at the top right corner of the screen. The content is then decided depending on whether you're in tutorial mode or not. And it should be translated from within transifex.
The rest of the tutorial content (the prompts that you see in the gray box in the middle of the imagery) has not changed yet, so it's still in firebase.

src/shared/views/LoadMore.js
laurentS: Yep, it's a [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) which the react docs recommend as one way to do conditional rendering.

src/shared/views/Login.js
laurentS: Not sure where `bal` came from, in practice it doesn't matter, since it's overwritten by the translated strings at runtime. I've removed it to make it clearer :)

src/shared/views/Mapper/index.js
laurentS: Same here, this is the help screen that is always the same.

src/shared/views/RecommendedCards.js
laurentS: Added.

Issue #377, opened 2020-08-18
Support basic http auth in imagery sources


laurentS:
Some imagery sources use a basic http auth scheme like `http://username:password@url` which the app should support.

laurentS:
Released in `2.0.6(0)`

PR #378, opened 2020-08-18
#377 Support basic http auth in imagery urls

laurentS:
This PR adds support for basic http auth in imagery urls.

When creating a project, nothing changes for regular imagery sources, but if a basic auth scheme is used, the url for the imagery must be entered in the form using the standard format: `http://username:password@url` (or https if required). There is no need to use base64 or any other encoding.

React native's http client does not handle this scheme by default, instead if requires passing a `Authorization: Basic username:password` header where `username:password` is base64 encoded.
I've updated the code that grabs the url for imagery to detect if the url follows this scheme (using a regular expression) and splits the url as needed to provide the required headers. The previous logic still applies to normal urls.

### DIFF
--- a/src/shared/views/Mapper/Tile.js
+++ b/src/shared/views/Mapper/Tile.js
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { firebaseConnect } from 'react-redux-firebase';
+import base64 from 'base-64';
 import {
     ImageBackground,
     View,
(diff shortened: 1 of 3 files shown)


src/shared/views/Mapper/Tile.js
```diff
@@ -150,14 +151,38 @@ export class _Tile extends React.Component<Props> {
...
+            return {
+                uri,
+                headers: {
+                    Authorization: `Basic ${base64.encode(
+                        `${username}:${password}`,
+                    )}`,
+                },
+            };
+        }
+        return { uri: rawUrl };
```
Hagellach37, line 172: this part is only reached, if the pattern don't match, right?

laurentS: yes, correct! The main reason I write it this way is because the linter enforces this pattern.

Issue #374, opened 2020-08-12
Crash upon app launch with v2.0.5(0)


laurentS:
**Describe the bug**
When launching the app on iPad, it crashes immediatly with the following sentry report:
```
TypeError:
undefined is not an object (evaluating 'S.supportedLanguages.filter(function(n){return n.code===u})[0].name')
  at render(src/shared/common/LanguageSelectionSplashScreen.js:124:11)
```

**App version**
`2.0.5(0)`

**To Reproduce**
Start the app. It seems to be related to data loading from app storage.

**Screenshots**
nothing, it just crashes at start

**Smartphone (please complete the following information):**
 - Device: iPad. Iphone seems ok

laurentS:
Details in sentry: https://sentry.io/organizations/missingmaps/issues/1836898431/?project=1326755&query=is%3Aunresolved#exception

laurentS:
It seems that this issue is not only happening on iPad, but on pretty much any platform. The upgrade path from 2.0.4 to 2.0.5 looks broken, so we need to release a 2.0.6 with a fix for this.

PR #379, opened 2020-08-18
#374 Fix crash on start due to xx language code

laurentS:
This PR attempts to fix the crash observed upon app start with `2.0.5(0)`, which appears to be caused by the `xx` language code that was the default value expected from redux. When upgrading from the previous version, it looks like the language code was set instead to `undefined`, which was not handled in the current code, leading to the crash.
This patch should fix this, and also handle the case where the code happens to be `xx`. This should not happen, but we might as well err on the side of safety at this point!

Sorry @Hagellach37 for not seeing the problem you had pointed out in your review of the PR for `2.0.5`!


Hagellach37 (approved):
looks good

PR #381, opened 2020-08-19
Prepare for release 2.0.6

laurentS:
This is mostly a hotfix release that fixes a crash when the app starts, which also seems to prevent the app from being released on the app store.
Also includes updated Dutch translations, and support for http basic auth in imagery.


jhenshall (approved):
Nice one for the fix @laurentS! Not a technical look-over but approving to get the fix moving.

PR #388, opened 2020-09-10
replace icons

jhenshall:
Removing licence text from icons – to be added to attributes screen instead.

@laurentS - I couldn't find the black version of the tap that seems to be used in the tutorial. Does the switch of the icon white to black happen in code? If not, point where I can add it before merging. Cheers!


laurentS:
I think I made the black version from the white one using an image editor. There does not seem to be a way to invert colours in code, unless you're prepared to write quite a lot of it.

PR #420, opened 2020-10-05
Prepare for release 2.0.8

laurentS:
This is a bugfix release:
- Added support for Russian language
- Updated some Czech translations
- Fixed a bug that prevented some users from completing the tutorial
- Minor dependency upgrades


src/shared/views/Mapper/CardBody.js
```diff
@@ -78,7 +78,7 @@ class _CardBody extends React.PureComponent<Props, State> {
         this.flatlist = null;
         this.scrollEnabled = !props.tutorial;
         this.tasksPerScreen = undefined;
```
Hagellach37, line 80: what does this line say? why do you set it to `undefined` here. 

don't we expect for the the "normal" project type, that tasks per screen is always 6?

src/shared/views/Mapper/CardBody.js
```diff
@@ -360,10 +360,20 @@ class _CardBody extends React.PureComponent<Props, State> {
...
         let result = 0;
         if (this.tasksPerScreen) {
-            result = this.tasksPerScreen[screenNumber].reduce(
-                (sum, task) => sum + task.referenceAnswer,
-                0,
-            );
+            console.log('tps', this.tasksPerScreen.length, screenNumber);
+            // $FlowFixMe
+            if (this.tasksPerScreen[screenNumber]) {
+                result = this.tasksPerScreen[screenNumber].reduce(
```
Hagellach37, line 366: I'm a bit confused here. I assume that `result` refers to the number of tasks expected for this screen, as it is defined by the sum of the reference answers.

Hagellach37 (approved):
Hey,
I can follow the things in general. And I think it makes sense to do it in this way. But still we need to learn more about the cause of the issue.

I guess that it goes somewhere in the direction of firebase data being loaded so quick enough, but still this would need further backing.

src/shared/views/Mapper/CardBody.js
laurentS: Yes, it's the total number of taps expected on the screen, if the user does a "perfect sequence". I had put that in the comment at the top of the function, let me know if I should clarify

src/shared/views/Mapper/CardBody.js
laurentS: I set it to `undefined` as a way to indicate that data hasn't loaded from the backend yet. Once it is loaded, `generateTasks()` fills it in with an array of `number of screens` items, each of which is 6 tasks long.

Issue #252, opened 2020-06-30
Add icons to mapping interactions


jhenshall:
Add icons to the mapping screen interactions to make it clearer what selection has been made. This would improve accessibility and help when colour filters are on (e.g. night mode).

User suggestion is to have an icon preferably on the tile in question showing the status of the tile in a darker form of the colour. So:

- Green tick
- yellow question mark
- red cross

Also, an icon at the next to the word of what you're looking for would be even better. So a building icon, and a road icon would work.

laurentS:
We've had 2 emails from colourblind users who've requested this sort of changes. I've invited them to comment on this ticket, but here are some excerpts from their messages:

> I'm slightly colorblind like many other males and the very minor difference in colors can be confusing. It's nice that i can count my clicks (1=building, 2 = maybe, 3= cloud) but the app would be much easier to use with some sort of other indicator.

and 

> as a colorblind user the shades that occur after the taps are basically indistinguishable for me. Would you please consider adding textures (like cross hatches), icons, and/or alternate color schemes? As the app currently is, I'm unable to use it. About 8% of males and 0.4% of females are colorblind, so improving accessibility would allow more people to use it.

As the ignorant developer, questions from me:
- would icons make more sense than hatches? From the previous comments, something like building/road/whatever we're looking for (see #374 ) | question mark | cloud?
- how big should they be? full tile, small icon in the corner?
- what colour should the icons be? I'm guessing something with high contrast?

Examples from around the web, or screenshots from apps would super helpful!

luisgregson:
An icon sounds like a good idea.

> would icons make more sense than hatches? From the previous comments, something like building/road/whatever we're looking for (see #374 ) | question mark | cloud?
how big should they be? full tile, small icon in the corner?
what colour should the icons be? I'm guessing something with high contrast?

I think it would be good to use numbers in a circle as the icon in the corner because the tutorial already has small number icons to indicate the number of times tapped, so it would be familiar to the user:
![image](https://user-images.githubusercontent.com/8599818/96725460-caa24e00-13a8-11eb-9182-580c3109dda6.png)

However, i've run a contrast test on that icon and it's only 2.45:1 which doesn't meet the WCAG AAA guidelines of 7:1
Maybe some icons like this might work:
![image](https://user-images.githubusercontent.com/8599818/96727611-279f0380-13ab-11eb-94b7-e484ac342c08.png)
[...]

PR #432, opened 2020-10-23
Update building footprint projects

laurentS:
This PR sets up building footprint projects to work with the WSF project requirements:
- support for google imagery
- updated design (to be adjusted)

This PR relates to the backend changes made in https://github.com/mapswipe/python-mapswipe-workers/pull/433

It is still WIP


src/shared/flow-types.js
```diff
@@ -5,12 +5,16 @@ import BottomProgress from './common/BottomProgress';
 
 // geographic types
 
+// a single point whose coordinates can be expressed in latitude/longitude
+// in geographic pixel coordinates or in image pixel coordinates
 export type Point = [number, number];
 
 export type Polygon = Array<Point>;
 
 export type BBOX = [number, number, number, number];
```
Hagellach37, line 14: could you add a comment here about which number refers to which positition? e.g. top left, right, east, west.... ;)

src/shared/flow-types.js
```diff
@@ -108,6 +112,7 @@ export type BuiltAreaTaskType = {
 
 // used only by projects of type BUILDING_FOOTPRINTS (type 2)
 export type BuildingFootprintTaskType = {
+    center: ?Point,
```
Hagellach37, line 115: does this mean that all BuildingFootprintTasks should have a `center` attribute? For the TMS based approach you don't need it at the moment I think.

What happens if the task will not have this attribute? (Just need to know for back end side)

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -1,43 +1,101 @@
...
+    PanResponderInstance,
+} from 'react-native/Libraries/Interaction/PanResponder';
 import { Path, Shape, Surface } from '@react-native-community/art';
 import tilebelt from '@mapbox/tilebelt';
 import type {
     BBOX,
     Point,
     Polygon,
-    ProjectType,
+    SingleImageryProjectType,
```
Hagellach37, line 15: where is this `SingleImageryProjectType` defined? I see it for the first time. What is it actually referring to? "normal" build area projects?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -1,43 +1,101 @@
...
+        this.panResponder = PanResponder.create({
+            onMoveShouldSetPanResponder: this.handleMoveShouldSetPanResponder,
+            onMoveShouldSetPanResponderCapture: this
+                .handleMoveShouldSetPanResponder,
+            onPanResponderRelease: this.handlePanResponderEnd,
+        });
+        this.imagerySize = 0;
+    }
+
+    // $FlowFixMe
```
Hagellach37, line 63: what is this comment about?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -56,7 +114,7 @@ export default class FootprintDisplay extends React.Component<Props> {
     };
 
     // return the center of the building footprint
-    getBuildingCentroid = (coords: Polygon): Point => {
+    getTaskGeometryCentroid = (coords: Polygon): Point => {
         const centroid: Point = coords
             .slice(0, -1)
             .reduce((acc, c) => [acc[0] + c[0], acc[1] + c[1]]);
```
Hagellach37, line 120: this is not needed for google imagery, but in general: what's your approach here?

"normally" you would derive the centroid of a polygon by computing it's bounding box and then you can get the center from these coordinates. So one need to find out max_lat, min_lat, max_lon, min_lon per geometry and from this get the centroid coordinates.

Hagellach37: what is `acc` and `c`?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -82,7 +140,59 @@ export default class FootprintDisplay extends React.Component<Props> {
         ];
     };
 
-    BBOXToCoords = (bbox: BBOX) => {
+    latLonZoomToPixelCoords = (lonLat: Point, zoom: number): Point => {
+        const [lon, lat] = lonLat;
```
Hagellach37, line 144: maybe we can add a comment here. e.g. this link with further information on what we mean by `PixelCoords`: https://docs.microsoft.com/en-us/bingmaps/articles/bing-maps-tile-system#pixel-coordinates

I think that now we still know, but maybe not in a few weeks, or if someone else will look at the code.

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -100,7 +210,7 @@ export default class FootprintDisplay extends React.Component<Props> {
         return [sw, nw, ne, se];
     };
 
-    getTileUrl = (tile: Tile) => {
+    getTileUrl = (tile: Tile): string => {
```
Hagellach37, line 213: just as a general question: this means that we can only rely on `Bing` if we do it with the TMS tiles?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -112,23 +222,88 @@ export default class FootprintDisplay extends React.Component<Props> {
...
+                            },
+                            styles.tileImg,
+                        ]}
+                        source={{ uri: tileUrl }}
+                    />
+                    <Surface
+                        height={GLOBAL.SCREEN_WIDTH}
+                        width={GLOBAL.SCREEN_WIDTH}
+                    >
+                        <Shape d={p} stroke="red" strokeWidth={1} />
```
Hagellach37, line 296: @jhenshall red or any other color?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -112,23 +222,88 @@ export default class FootprintDisplay extends React.Component<Props> {
...
+                    </Surface>
+                </View>
+            );
+        }
+        // all other imagery sources work with 4 tiles shown at the same time
```
Hagellach37, line 301: don't we need an if else statement thing here? I see that you used `if` above, but there is no `else`. not needed?

src/shared/views/BuildingFootprint/Validator.js
```diff
@@ -2,145 +2,218 @@
...
+    longNarrowButtonText: {
+        color: COLOR_WHITE,
+        fontWeight: 'bold',
+    },
 });
 
-const FOOTPRINT_CORRECT = 1;
-const FOOTPRINT_NEEDS_ADJUSTMENT = 2;
-const FOOTPRINT_NO_BUILDING = 3;
+const WSF_NO_BUILDING = 0;
```
Hagellach37, line 63: I think you could have this more generic:

FOOTPRINT_NO = 0
FOOTPRINT_YES = 1
FOOTPRINT_MAYBE = 2
FOOTPRINT_BAD = 3

For footprint we will (for now) always have a yes-no question. So the above will be valid for all projects, and no need to do it WSF specific here.

Hagellach37 (approved):
Nice. Really cool. I will now give it a try in the app. :)

Most comments are only of general character to help me understand what's going on. We might need to review some of the parts related to the TMS based loading, but this can be out of scope for WSF and we should address this at a later stage.

But maybe we can document in an issue that we keep track of it. :)

src/shared/flow-types.js
laurentS: The question mark in `flow` is like `Optional[number]` in python type annotations. So it can be missing and the app is still happy.

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: It's defined in `flow-types.js`. Basically, it means "not Change Detection" I think, so projects that only use a single source of imagery. The way it is defined now, Completeness projects would be single imagery, which is weird. We should update the definitions for those.

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: This tells `flow` to ignore the next line because I couldn't find the correct type definition for the event. I'll look a bit harder to fix it.

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: Ok, so my definition of centroid might be wrong then. I basically take the average of all latitude and same with longitudes, and that gives me a point which I guess is the centre of gravity of the geometry, probably not exactly the same as centroid. But I guess for this purpose, it's fine, it helps us put the shape roughly in the centre of the image.

laurentS: oh and `acc` is "accumulator" and `c` is just short for "coordinate" I think. See [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) for more details.

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: good call. Comment added with your link

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: for now, yes. I think we should refactor the code to have a small imagery url interpolation library shared across all project types. https://github.com/mapswipe/mapswipe/issues/433

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: This is probably a question to check with regards to accessibility. I guess an option could be 3 lines white/black/white to have a very strong contrast. See #252

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: Just following eslint's preferred format, which I've used everywhere in the code at this point. It's arguable whether it's the best style or not, but I'd rather just follow a linter, less brainpower required :)

src/shared/views/BuildingFootprint/Validator.js
laurentS: fixed!

src/shared/flow-types.js
laurentS: done!

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: Actually, I'm wrong, it's not centre of gravity of the shape, more like the centre of gravity of the points...

Hagellach37: jo, I will create an issue for this okay? :) It's not urgent now and a good thing to fix for interested students of geography. ;)

src/shared/views/BuildingFootprint/FootprintDisplay.js
Hagellach37: #435

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -78,14 +115,44 @@ export default class FootprintDisplay extends React.Component<Props> {
...
+                useNativeDriver: false,
+            }),
+        ]).start();
+    };
+
     handlePanResponderEnd = (event: PressEvent, gestureState: GestureState) => {
         // swipe completed, decide what to do
         const { nextTask, previousTask } = this.props;
         const swipeMinLength = 0.2;
         if (gestureState.dx < -GLOBAL.TILE_VIEW_HEIGHT * swipeMinLength) {
```
Hagellach37, line 146: coudl you add a comment here? I think it defines the threshold for swiping?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -96,9 +163,9 @@ export default class FootprintDisplay extends React.Component<Props> {
         const [minLon, minLat, maxLon, maxLat] = screenBBox;
         // geographic coords to screen pixels
         const lon2x = (lon) =>
-            ((lon - minLon) / (maxLon - minLon)) * this.imagerySize;
+            ((lon - minLon) / (maxLon - minLon)) * this.imageryHeight;
```
Hagellach37, line 166: should it be GLOBAL_SCREEN_WIDTH here instead of `imageryHeight`?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -290,50 +385,50 @@ export default class FootprintDisplay extends React.Component<Props> {
...
-            zoomLevel,
+            this.zoomLevel,
         );
-        const tiles = this.getTilesFromScreenCorners(corners, zoomLevel);
+        const tiles = this.getTilesFromScreenCorners(corners, this.zoomLevel);
         const tileUrls = tiles.map(this.getTileUrl);
 
-        const shiftX = (swCornerTile[0] % 1) * this.imagerySize;
-        const shiftY = (swCornerTile[1] % 1) * this.imagerySize;
+        const shiftX = (swCornerTile[0] % 1) * this.imageryHeight;
```
Hagellach37, line 414: here again, do we have to use GLOBAL_SCREEN_WIDTH instead?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -290,50 +385,50 @@ export default class FootprintDisplay extends React.Component<Props> {
...
+        const shiftY = (swCornerTile[1] % 1) * this.imageryHeight;
         return (
             <View
                 {...this.panResponder.panHandlers}
                 style={{
-                    height: this.imagerySize,
+                    height: this.imageryHeight,
                     overflow: 'hidden',
-                    width: this.imagerySize,
+                    width: this.imageryHeight,
```
Hagellach37, line 422: see above

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -290,50 +385,50 @@ export default class FootprintDisplay extends React.Component<Props> {
...
             >
                 <View
                     style={{
                         position: 'absolute',
                         left: -shiftX,
                         top: -shiftY,
-                        height: this.imagerySize * 2,
-                        width: this.imagerySize * 2,
+                        height: this.imageryHeight * 2,
+                        width: this.imageryHeight * 2,
```
Hagellach37, line 431: see above?

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -349,7 +444,7 @@ export default class FootprintDisplay extends React.Component<Props> {
                     <Image
                         style={[
                             {
-                                left: this.imagerySize,
+                                left: this.imageryHeight,
```
Hagellach37, line 447: ...

src/shared/views/BuildingFootprint/Validator.js
```diff
@@ -141,76 +164,135 @@ class _Validator extends React.Component<Props, State> {
...
         const { currentTaskIndex } = this.state;
         if (result !== null && result !== undefined) {
+            // the user tapped a button, save result and update how far
+            // we've worked
             submitResult(result, this.expandedTasks[currentTaskIndex].taskId);
-            this.tasksDone = currentTaskIndex;
+            if (currentTaskIndex > this.tasksDone) {
+                this.tasksDone = currentTaskIndex;
+            }
         } else if (currentTaskIndex > this.tasksDone) {
```
Hagellach37, line 173: I don't get this line. it looks exactly the same as you have in line 170

Hagellach37 (commented):
hey,
most things are clear. Could you just double check the usage of  `imageryHeight` and `width`?

src/shared/views/BuildingFootprint/Validator.js
laurentS: Correct. But the one above is under another condition. The first one is when the user taps a button, the second one when they swipe.

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: Added!

src/shared/common/ScaleBar.js
```diff
@@ -6,12 +6,19 @@ import { Path, Shape, Surface, Text } from '@react-native-community/art';
 import GLOBAL from '../Globals';
 
 type Props = {
+    // whether to shift the scale bar a little bit up from the bottom
+    alignToBottom: boolean,
```
Hagellach37, line 10: when would this not be shifted a little bit? for "normal" projects?

src/shared/common/ScaleBar.js
```diff
@@ -6,12 +6,19 @@ import { Path, Shape, Surface, Text } from '@react-native-community/art';
...
+    // using the tile size. This is useful for building footprint projects
+    // with TMS imagery as we use a few tricks there to cover the entire screen
+    // with imagery
+    useScreenWidth: boolean,
     visible: boolean,
     zoomLevel: number,
 };
 
-const getScaleBar = (meters, feet, tileWidth) => {
+const getScaleBar = (meters, feet, tileWidth, referenceSize) => {
```
Hagellach37, line 21: just to understand better: for TMS projects there will be referenceSize = tileWidth, but for Building Footprint this will be two different values

src/shared/common/ScaleBar.js
```diff
@@ -70,19 +83,20 @@ export default (props: Props) => {
             break;
     }
 
-    const p = getScaleBar(meters, feet, tileWidth);
+    const referenceSize = useScreenWidth
+        ? GLOBAL.SCREEN_WIDTH
+        : GLOBAL.TILE_SIZE;
```
Hagellach37, line 88: okay I think this relates to my question above

src/shared/flow-types.js
```diff
@@ -5,12 +5,35 @@ import BottomProgress from './common/BottomProgress';
...
+// coordinates expressed in pixel coordinates (x, y, from "top left" of the globe)
+export type PixelCoordsX = number;
+export type PixelCoordsY = number;
+export type PixelCoordsPoint = [PixelCoordsX, PixelCoordsY];
+
+// coordinates in "image coords", ie: measured in pixels from the
+// top left of a tile, as [x, y]
+export type ImageCoordsPoint = [number, number];
+
+// zom levels are always integers (typically 1 to 20-21)
```
Hagellach37, line 25: zoom

src/shared/views/BuildingFootprint/FootprintDisplay.js
```diff
@@ -1,20 +1,37 @@
...
     BuildingFootprintTaskType,
     Tile,
+    ZoomLevel,
 } from '../../flow-types';
 
 const GLOBAL = require('../../Globals');
 
-const tileSize = GLOBAL.SCREEN_WIDTH;
+// tileSize is only used for tile based imagery (ie: everything but google)
+const tileSize = GLOBAL.SCREEN_WIDTH / 1;
```
Hagellach37, line 34: why do you divide by 1?

src/shared/views/BuildingFootprint/Validator.js
```diff
@@ -2,145 +2,300 @@
...
+    longNarrowButtonText: {
+        color: COLOR_WHITE,
+        fontWeight: 'bold',
     },
 });
 
-const FOOTPRINT_CORRECT = 1;
-const FOOTPRINT_NEEDS_ADJUSTMENT = 2;
-const FOOTPRINT_NO_BUILDING = 3;
+const FOOTPRINT_NO_BUILDING = 0;
```
Hagellach37, line 82: I would rename this to FOOTPRINT_YES, FOOTPRINT_NO

src/shared/views/BuildingFootprint/Validator.js
```diff
@@ -2,145 +2,300 @@
...
-        const { currentTaskId } = this.state;
-        if (!group.tasks) {
+        const { project, results } = this.props;
+        const { currentTaskIndex } = this.state;
+        if (!this.expandedTasks) {
             return <LoadingIcon />;
         }
-        const currentTask = group.tasks.find((t) => t.taskId === currentTaskId);
+        const currentTask = this.expandedTasks[currentTaskIndex];
+        // if tasks have a center attribute, we know they're grouped by 9
```
Hagellach37, line 210: maybe for clarification: for which tasks do we expect a center attribute? only for the google ones?

Hagellach37 (approved):
Wow this was a lot of work. :)

You did a really good job on the comments I think.

src/shared/common/ScaleBar.js
laurentS: I don't shift the scale bar on building footprint with TMS imagery. it places the scale bar much too high in the image, which looks weird. I'll spare you the details of why it happens. I couldn't figure out an obvious way to unify the code, so that's what we have for now :)

src/shared/common/ScaleBar.js
laurentS: Yup. I think the comment about `useScreenWidth` at the top and the code below should make it clear. If not, just comment back here, and I'll try to elaborate.

src/shared/common/ScaleBar.js
laurentS: yup :)

src/shared/flow-types.js
laurentS: fixed.

src/shared/views/BuildingFootprint/FootprintDisplay.js
laurentS: oops, probably some leftovers from when I was trying to adjust sizes and stuff. I guess there was a 2 there at some point.

src/shared/views/BuildingFootprint/Validator.js
laurentS: makes sense. fixed.

src/shared/views/BuildingFootprint/Validator.js
laurentS: Yes, for now, we only use it with google imagery, if it exists (it's not required, it's just possible to add it for optimisation). I've added a bit of explanation in the code, hopefully it helps us remember what we did here :)

src/shared/views/ChangeDetection/Task.js
```diff
@@ -159,6 +159,8 @@ export default class ChangeDetectionTask extends React.PureComponent<
                             icon: 'fixme',
                         }}
                         boxType="fixme"
+                        bottomOffset="45%"
```
Hagellach37, line 162: not sure why you changed something in change detection here.

src/shared/views/Mapper/CardBody.js
```diff
@@ -588,6 +590,8 @@ class _CardBody extends React.PureComponent<Props, State> {
                         <TutorialBox
                             content={tutorialContent}
                             boxType={tutorialMode}
+                            bottomOffset="80%"
```
Hagellach37, line 593: again here you changed the placement of the tutorial box.

ahh, probably before you set it globally and now per project type?

Hagellach37 (approved):
looks good to me. :+1:

src/shared/views/Mapper/CardBody.js
laurentS: Yes, the value was hardcoded in the `TutorialBox` component, but I added a prop on the component so we can reuse it between different project types. The original placement is the same, I've just changed where the value comes from.

src/shared/views/ChangeDetection/Task.js
laurentS: see below

PR #492, opened 2021-05-10
Draft: Fix change detection imagery

Hagellach37:


laurentS (changes requested):
It looks like you've made this PR off of `master`, which we should avoid if possible. Ideally, you'd branch off of `dev`, commit to that branch, then do a PR from there onto `dev`, to follow the process we had defined in [this doc page](https://github.com/mapswipe/mapswipe/blob/master/docs/workflow.md#workflow-overview-for-the-core-team)

PR #495, opened 2021-05-10
add function to calculate best zoom level for task geometry

### LINKED ISSUE #494: Large polygons for Building Footprint type are not correctly displayed
**Describe the bug**
The building footprint type uses zoom level 19 as a default value. This fits for rather small polygons such as buildings, but doesn't display things correct if the polygons depict larger objects, e.g. entire settlements or villages.

The zoom level should be set based on the extent of the object.

**App version**
both, all versions

**To Reproduce**
Please provide steps to reproduce the problem:
1. Map the project "Validation Missing Millions - Uganda" in the dev version of the app.
 
**Screenshots**

* this is how it looks: the object is too large and the boundary is beyond what is shown on the screen at zoom level 19

* this is how it should look like: object displayed at zoom level 16

Hagellach37:
Add a function to be able to display objects of various size in the footprint project type.

closes #494 

Somethings to discuss / where I'm not 100% sure:

* where/when should the zoomLevel be defined? It used to be defined in `constructor()`, but there we just set it to a hard coded value
* so where is the best place to call the new function `getZoomLevelFromCoords()`?
* I'm not getting when a function like `componentDidUpdate()` is called and the way I have implemented it at the moment the zoomLevel will be calculated more than once per task, which might not be ideal


laurentS:
>     * where/when should the zoomLevel be defined? It used to be defined in `constructor()`, but there we just set it to a hard coded value

The way you've done it now, the zoom level is calculated twice for each task (at least). While it's not a crazy expensive calculation, it's always nice if we can save some CPU cycles :)

>     * so where is the best place to call the new function `getZoomLevelFromCoords()`?
I'd say as late as possible before the imagery is loaded, because that's the expensive bit. Since there is a preloading mechanism, and you're dealing with multiple tasks in parallel (2 really), maybe you could create an array of zoom levels, initially filled with `undefined` and when you preload the imagery for a task, calculate the zoom level, store it in the array, so that you can then retrieve it when actually loading the imagery. Benefit would be that you don't need to recalculate anything if the user swipes back, etc... Since preloading is not guaranteed, I'd check the value when using it, and if it's `undefined` then calculate/save at that time.

>     * I'm not getting when a function like `componentDidUpdate()` is called and the way I have implemented it at the moment the zoomLevel will be calculated more than once per task, which might not be ideal

`componentDidUpdate()` is a react method on all components which is called when props for the component are updated, for instance each time the `task` prop is changed, which is pretty much what you want here.

Hagellach37:
Hey,
I thought a bit about your suggestions and also tried around a bit, but are still not really sure how to do this properly. I just changed the code slightly to pass zoom as an argument to `getGoogleImageryUrl()` and `getTMSImageryUrls()`. So we need to calculate the zoom only once per task.

As I understand it the preloading of the images doesn't do anything except downloading the images. So when the next task is loaded, we can get these images from cache. But in the existing code there is no example how a pre calculated value is passed. (E.g. we do NOT calculate the path for the prefetched task geometry in the footprint project type.)

And to be honest, I'm not even sure if the prefetching of the images is actually working. I added a log to check. Any ideas why this would be the case?

laurentS:
> As I understand it the preloading of the images doesn't do anything except downloading the images. So when the next task is loaded, we can get these images from cache. But in the existing code there is no example how a pre calculated value is passed. (E.g. we do NOT calculate the path for the prefetched task geometry in the footprint project type.)

Yes, that's basically all it does. The code doesn't pass around the precalculated url, it actually recalculates it (something we could prevent as well!) but what happens is that we get the same url, and when we pass it to the image component, the OS sees that this image is already cached and shows it directly from there, instead of redownloading it.
 
> And to be honest, I'm not even sure if the prefetching of the images is actually working. I added a log to check. Any ideas why this would be the case?

Hmm, that's embarrassing. I thought we had tested it quite a bit and were happy with the results. Are you on a super fast connection?

Hagellach37:
Hey @laurentS ,
I'm not sure what's the best way forward here. Do we want to avoid calling the function twice or is it more "nice to have"? :)

PR #496, opened 2021-05-10
display imagery correctly and avoid overlay in change detection projects

Hagellach37:
This is just a tiny change to how the images are loaded for the change detection project type. (second try)

closes #492

### DIFF
--- a/src/shared/common/SatImage.js
+++ b/src/shared/common/SatImage.js
@@ -105,6 +105,7 @@ export default class SatImage extends React.Component<Props, State> {
                         results={0}
                         style={styles.imageBackground}
                         tile={task}
+                        source={source}
                         tutorial={false}
                     />
                 ) : (
--- a/src/shared/views/ChangeDetection/Tile.js
+++ b/src/shared/views/ChangeDetection/Tile.js
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import { firebaseConnect } from 'react-redux-firebase';
 import {
     ImageBackground,
+    Image,
     View,
     StyleSheet,
     TouchableHighlight,
@@ -52,6 +53,7 @@ type Props = {
     onToggleTile: (ResultType) => void,
     results: number,
     style: ViewStyleProp,
+    source: Image.ImageSourcePropType,
     tutorial: boolean,
 };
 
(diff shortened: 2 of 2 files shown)


src/shared/common/SatImage.js
```diff
@@ -105,6 +105,7 @@ export default class SatImage extends React.Component<Props, State> {
                         results={0}
                         style={styles.imageBackground}
                         tile={task}
+                        source={source}
```
laurentS, line 108: Since this is a shared component, did you check that adding this prop has no weird side effects on the other project types?

src/shared/views/ChangeDetection/Tile.js
```diff
@@ -199,38 +179,14 @@ export class _Tile extends React.PureComponent<Props> {
...
-                        styles.tileOverlay,
-                        { backgroundColor: overlayColor },
-                    ]}
-                    key={`view-${taskId}`}
-                >
-                    {animatedRows}
-                </View>
-            );
-        }
+        const comp = (
```
laurentS, line 182: Is it worth having this variable here, since it always has the same value? Maybe inline this component into the main one a few lines down for clarity?

laurentS (approved):
Looks alright to me :)

src/shared/views/ChangeDetection/Tile.js
Hagellach37: Hey, yes good comment. I moved it down.

src/shared/common/SatImage.js
Hagellach37: So, I mapped a normal project. But to be honest didn't have a proper way to test for things. What would you suggest?

For the other project types, e.g. in the normal one, the `source` attribute is not used. So unless there are side effects when you don't pass an argument, then I think we are fine.

laurentS: Not much else I think. if the usual projects work, the tutorials as well, then it should be fine. Oh and `eslint / flow` are happy as well, so all good I think.

Hagellach37: okay, yeah also checked the tutorial.

Hagellach37:
Hey @laurentS thanks for taking a look at this. I'm going to merge now. :) :rocket:

PR #501, opened 2021-05-31
Setup GitHub actions

### LINKED ISSUE #445: Switch CI from travis
We currently use travis for our CI builds, but travis has become painfully slow to start builds in the last month or 2.
The latest builds wait in a queue for almost an hour before starting (the android build at least, iOS builds don't seem to be affected). Once started, they build roughly like before.
While it's not a disaster, it slows down the overall process. I've noticed the same slowing down on other projects, so it might be that travis is cutting resources for opensource/free users.
It might be worth exploring switching to github actions in order to speed things up again.

laurentS:
This PR converts our continuous integration/deployment pipeline from travis to github actions.

Functionality should be the same as before:
- CI runs on all pushes to github and checks lint/flow and builds both android and iOS packages
- on tagged commits, it releases either the beta or production version as before

I've tried to improve documentation a bit as well as maintainability. Please comment ad lib on these aspects, and ask any clarifying questions if anything is not obvious.

Fixes #445.


.github/workflows/build.yml
```diff
@@ -0,0 +1,200 @@
...
+          SENTRY_PROPERTIES: ${{ secrets.FILE_SENTRY_PROPERTIES }}
+        run: echo "$SENTRY_PROPERTIES" | base64 -d > ./android/sentry.properties
+      - name: Decode development google-services.json
+        env:
+          FILE_ANDROID_GOOGLE_SERVICES_DEV: ${{ secrets.FILE_ANDROID_GOOGLE_SERVICES_DEV }}
+        run: echo "$FILE_ANDROID_GOOGLE_SERVICES_DEV" | base64 -d > ./android/app/src/dev/google-services.json
+      - name: Decode production google-services.json
+        env:
+          FILE_ANDROID_GOOGLE_SERVICES_PRODUCTION: ${{ secrets.FILE_ANDROID_GOOGLE_SERVICES_PRODUCTION }}
+        run: mkdir -p ./android/app/src/production ; echo "$FILE_ANDROID_GOOGLE_SERVICES_PRODUCTION" | base64 -d > ./android/app/src/production/google-services.json
```
Hagellach37, line 63: here you first create the directory. For the dev service account you don't do this.  Is there a reason why you have do it here?

.github/workflows/build.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: Increase filesystem inotify watch limit
+        run: sudo sysctl fs.inotify.max_user_watches=524288
+      - name: Build dev APK
+        env:
+          ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_KEY_ALIAS: ${{ secrets.ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_KEY_ALIAS }}
+          ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_KEY_PASSWORD: ${{ secrets.ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_KEY_PASSWORD }}
+          ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_STORE_FILE: ${{ secrets.ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_STORE_FILE }}
+          ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_STORE_PASSWORD: ${{ secrets.ORG_GRADLE_PROJECT_MAPSWIPE_RELEASE_STORE_PASSWORD }}
+        run: yarn buildDev
+      - name: Build production APK
```
Hagellach37, line 102: Hey,
so just checking here: We will always build the production APK for every commit? But only upload if it is tagged, but not when the release is tagged with beta?

.github/workflows/build.yml
```diff
@@ -0,0 +1,200 @@
...
+        if: startsWith(github.ref, 'refs/tags/') && !endsWith(github.ref, '-beta')
+        with:
+          artifacts: "./app/build/outputs/apk/production/release/app-production-release.apk"
+          prerelease: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Release to PlayStore
+        uses: maierj/fastlane-action@v2.0.1
+        # only deploy a release on a non-beta tag
+        # FIXME: there does not seem to be a way with github actions (as of may 2021) to get
+        # both tag and branch, so for now we will rely on the user not tagging randomly
```
Hagellach37, line 128: hey,
this looks interesting. So basically we could do a production release from any branch and it will be released to playstore.

What does this mean?

When it's released to playstore there is still manual approval needed, to make it available to users? So even if somebody would accidentally release a production version this would not be a problem, since we could just delete it in playstore?

.github/workflows/build.yml
```diff
@@ -0,0 +1,200 @@
...
+        run: |
+          gpg --quiet --batch --yes --decrypt --passphrase="$FILE_DECRYPTION_PASSPHRASE" --output ./ios/cfg/mapswipe.dev_at_gmail_rsa_key_for_travis_ci ./ios/cfg/mapswipe.dev_at_gmail_rsa_key_for_travis_ci.gpg
+          chmod 600 ./ios/cfg/mapswipe.dev_at_gmail_rsa_key_for_travis_ci
+      - name: Add gitlab to ssh_known_hosts
+        env:
+          SSH_KNOWN_HOSTS: ${{ secrets.SSH_KNOWN_HOSTS }}
+        run: echo "$SSH_KNOWN_HOSTS" > ~/.ssh/known_hosts
+      - name: Build the app and run unit tests
+        # The test lane refreshes the development certificates, compiles the app and launches the unit tests.
+        # Actual Tests have been disabled since Travis has issues launching the simulator, but the app is being compiled.
```
Hagellach37, line 182: Here you mention Travis, but it should be Github Actions?

Hagellach37 (commented):
Hey @laurentS 
I checked the code and most things look good to me.

I'm still thinking about how we can handle it to only build production releases when needed? Currently we just always build everything and always release either production or beta when commits are tagged.

As you mentioned there seems to be no way to get the `tag` and `branch` at the same time. Can you explain more why you think it is like this? Maybe not the shortest solution, but it looks like it would be possible to define separate workflows for this: https://stackoverflow.com/questions/58139406/only-run-job-on-specific-branch-with-github-actions

So instead of having one workflow that does everything. We could split things up into several workflows. For example having a workflow for:
- android production release
- android dev release
- ios production release
- ios dev release

Is this an option or are there other reasons why we would not do it like this?

.github/workflows/build.yml
laurentS: The `dev` directory already exists because there are files in it (icons/splash screen). I'm not sure why there are no equivalent files in the production folder, but that's how it is for now, and it works, so I won't wake up the sleepy bear :bear:

.github/workflows/build.yml
laurentS: yes, this is what we were doing in travis. For now, I focused on getting the same functionality up and running, but I think we should split things up as you suggested elsewhere, it feels a bit silly to build the app over and over each time.

.github/workflows/build.yml
laurentS: I'm adding a check in the `scripts/version.sh` script to select `beta` or `production` based on the current branch. Obviously this does not prevent manual tagging which would throw things off, but as you say, we still have a manual step in the app stores after this, and there is only a limited number of people who can actually trigger this job, so I think overall the risk is low.

.github/workflows/build.yml
laurentS: the magic of copy/pasting! thanks for picking it up!

laurentS:
> I'm still thinking about how we can handle it to only build production releases when needed? Currently we just always build everything and always release either production or beta when commits are tagged.
> 
> As you mentioned there seems to be no way to get the `tag` and `branch` at the same time. Can you explain more why you think it is like this? Maybe not the shortest solution, but it looks like it would be possible to define separate workflows for this: https://stackoverflow.com/questions/58139406/only-run-job-on-specific-branch-with-github-actions

I don't know why they implemented it like this, to me it's a bad design because it would be easy for github to add a variable that has this value. That said, when I tagged a commit just now, it triggered 2 pipelines (like it used to do in travis): 1 for the commit, 1 for the tag.

> So instead of having one workflow that does everything. We could split things up into several workflows. For example having a workflow for:
> 
>     * android production release
>     * android dev release
>     * ios production release
>     * ios dev release
> 
> Is this an option or are there other reasons why we would not do it like this?

Yes, this is what I'd like to move towards (once previous functionality is working) in order to make runs faster, and also in a way reduce resource usage (CI sometimes feels like a magic way to burn energy for nothing).
I also want to avoid having 4 different workflows with 90% code reuse, but I think there must be a way to share the code between the workflows, I'll look into it.

jhenshall (approved):
Nicely done man! A lot gone into this and great to get it done.

PR #519, opened 2021-06-12
Setup GitHub actions (move from Travis)

laurentS:
This PR implements CI on github actions to replace the travis setup. See #445 for details about the rationale behind it.

Opening the PR mostly to be able to keep notes about the transition

This is a follow up to #501


laurentS:
Keeping some notes here about the issues we've seen with running the iOS build from testflight.
- Initial version 2.0.11(11) crashed immediately after starting. When switching over from travis to github actions, we implicitly moved from macos 10.14.6 to 10.15.7 and Xcode from 11.3 to 12.4. The app crashed immediately when run from testflight, and while no exact bug report existed on github, all the similar ones seem to point at version compatibility issues between react-native and macos/Xcode. Hopefully this fixes it.
- 2.0.11(12) tried building using the same xcode version as we last had in travis (Xcode 11.3.1), but the upload to testflight was rejected because it only accepts builds done with SDK 14+.
- `Starting April 26, 2021, all iOS and iPadOS apps submitted to the App Store must be built with Xcode 12 and the iOS 14 SDK.` from https://developer.apple.com/news/?id=ib31uj1j so we can't revert to an older Xcode version.
- which means we most likely have to upgrade react-native to get around this :( Looking for confirmation in RN release notes. Hopefully there is a workaround...

It's worth noting that there is a long thread about building errors on Xcode 12.5, which we're not on yet, but I guess we'll face those in due course :(

.github/workflows/ios.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     runs-on: macOS-latest
     steps:
       - uses: actions/checkout@v1
+      - name: List available versions of Xcode
```
Hagellach37, line 15: not sure if we need that step still

Hagellach37 (approved):
Hey @laurentS 
looks good to me. :) Do we still need to log available xcode versions?

.github/workflows/ios.yml
laurentS: Not really, but I'm tempted to keep it since it takes virtually no time, and it might hint at upcoming changes when xocde 12.5 turns up on github actions, as it looks like this will cause a bunch of upgrade issues once more :(

Hagellach37: okay, so let's keep this line then.

PR #520, opened 2021-06-17
add translation for change detection before and after

Hagellach37:


laurentS (approved):
looks alright! the real test is seeing whether transifex accepts your changes when they're merged into `dev` :)

Issue #502, opened 2021-06-02
Make apple releases work again in CI


laurentS:
In #501 I've setup github actions to build and deploy the app as was done before in travis.

As I write this, we are waiting for access to Apple's API connection system as the system we used before is now disabled (they keep upgrading their security, but make it really hard for developers to follow).

We need to:
- get a connect API key https://docs.fastlane.tools/app-store-connect-api/#creating-an-app-store-connect-api-key
- use it in our CI code
- verify that pushing to testflight and appstore works

laurentS:
Fixed with #524

Issue #405, opened 2020-09-16
Upgrade react-native to 0.63


laurentS:
We're now 2 "major" versions behind the current RN. While it's not problematic yet, we should avoid falling too far down behind, as detailed in #61 and related tickets.
This ticket requires an upgrade to 0.62 first... :(

Effort is only an estimate, hopefully the upgrade is smooth!

laurentS:
There are 2 (1 minor, 1 medium) security vulnerabilities that should be fixed by upgrading from RN 0.61.5 (current). They are likely not affecting the app from a quick assessment, but it's one more argument for upgrading as soon as possible.

Issue #216, opened 2020-05-26
Upgrade RNFirebase to v6.x.x


laurentS:
As of May 26, 2020 (and probably a bit before), rnfirebase v5.x.x is deprecated and unsupported.

We should upgrade to v6 soon. According to the [migration guide](https://rnfirebase.io/migrating-to-v6), it does not seem to require major changes to the code, except for a few search and replace changes.
Check that react-redux-firebase supports this new version.

This is probably done best as a separate update/release to help isolate the cause of potential problems.

laurentS:
closed in #544

PR #524, opened 2021-07-01
Upgrade react-native to 0.64.2

laurentS:
This PR is fairly messy and is a continuation of #519 (which should be merged before this one).

While moving to github actions, we ended up hitting a problem with uploading to testflight (see [my other comment](https://github.com/mapswipe/mapswipe/pull/519#issuecomment-860061320)) which more or less forced us to upgrade:
- react-native to 0.64.2 (the latest at the time)
- react-native-firebase to v6+ as the previous versions were not supported anymore
- a few other dependencies

This PR tracks these upgrades, and multiple attempts to fix a crash on app launch on iOS which was due to an environment variable missing in CI for iOS :roll_eyes: 

Issues covered by this PR:
- #405
- #216 
- #502


.flowconfig
```diff
@@ -37,24 +34,22 @@ emoji=true
 esproposal.optional_chaining=enable
 esproposal.nullish_coalescing=enable
 
+exact_by_default=true
```
Hagellach37, line 37: that's how we do it. by default, also exact. :D

but for curiosity: what does it mean?

.github/workflows/android.yml
```diff
@@ -12,26 +12,33 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
+
       - uses: actions/setup-node@v1
         with:
-          node-version: '16.1.0'
+          node-version: '14.16.1'
```
Hagellach37, line 18: could we here move again up to 16.1.0? or are there any other reason to use v14?

.github/workflows/ios.yml
```diff
@@ -12,69 +12,129 @@ jobs:
...
     steps:
       - uses: actions/checkout@v1
+
+      - name: List available versions of Xcode
+        run: ls -n /Applications/ | grep Xcode*
+
       - uses: actions/setup-node@v1
         with:
-          node-version: '16.1.0'
+          node-version: '14.16.1'
```
Hagellach37, line 21: same question as above

.github/workflows/ios.yml
```diff
@@ -12,69 +12,129 @@ jobs:
...
         with:
           ruby-version: '2.7.3'
+
       - name: Install Fastlane
         run: bundle install
+
       - name: Install cocoapods
         run: gem install xcpretty cocoapods
+
+      - name: TEMP manually run pod install
```
Hagellach37, line 36: Does `TEMP` indicate here that you actually planned to remove these steps?

.github/workflows/ios.yml
```diff
@@ -12,69 +12,129 @@ jobs:
...
           MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
         run: cd ios && bundle exec fastlane release
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: build.log
+          path: ~/Library/Logs/gym/*.log
+
+      - name: Find output files
```
Hagellach37, line 128: do we still need this? looks like you tried to log which files there are

android/app/src/main/AndroidManifest.xml
```diff
@@ -16,22 +16,14 @@
...
+        android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
+        android:launchMode="singleTask"
         android:windowSoftInputMode="adjustResize">
         <intent-filter>
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
         </intent-filter>
       </activity>
-      <activity android:name="com.facebook.react.devsupport.DevSettingsActivity" />
-      <service android:name="io.invertase.firebase.messaging.RNFirebaseMessagingService">
```
Hagellach37, line 29: I think we've said this somewhere else already. but this means that we can't send notifications, right?

android/app/src/main/java/org/missingmaps/mapswipe/MainActivity.java
```diff
@@ -5,7 +5,6 @@
 import com.facebook.react.ReactActivityDelegate;
 import com.facebook.react.ReactRootView;
 import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;
-import org.devio.rn.splashscreen.SplashScreen;
```
Hagellach37, line 8: do we want to have the splashscreen again? Or is there any other reason, why we should remove it?

android/app/src/main/java/org/missingmaps/mapswipe/MainApplication.java
```diff
@@ -6,20 +6,10 @@
 import com.facebook.react.PackageList;
 
 import com.facebook.react.ReactApplication;
-import io.invertase.firebase.RNFirebasePackage;
+import com.facebook.react.ReactInstanceManager;
 import com.BV.LinearGradient.LinearGradientPackage;
 import com.reactnativecommunity.asyncstorage.AsyncStoragePackage;
-import io.sentry.RNSentryPackage;
```
Hagellach37, line 12: did you remove sentry as well? or do we still get the errors reports in sentry?

android/app/src/main/java/org/missingmaps/mapswipe/MainApplication.java
```diff
@@ -6,20 +6,10 @@
...
-import com.reactnativecommunity.viewpager.RNCViewPagerPackage;
 import com.reactnativecommunity.netinfo.NetInfoPackage;
-import io.invertase.firebase.RNFirebasePackage;
```
Hagellach37, line 15: all these packages have been removed due to the upgrade of the firebase package? and are now covered by something else?

ios/Podfile
```diff
@@ -1,72 +1,36 @@
...
 
   target 'mapswipeTests' do
-    inherit! :search_paths
+    inherit! :complete
     # Pods for testing
   end
 
-  use_native_modules!
-end
+  # Enables Flipper.
```
Hagellach37, line 26: becomes it has come up so often: What is the flipper?

src/shared/views/Welcome.js
```diff
@@ -68,14 +67,14 @@ class _WelcomeScreen extends React.Component<Props> {
         if (welcomeCompleted) {
             this.finishWelcomeScreens();
         } else {
-            SplashScreen.hide();
+            // SplashScreen.hide();
```
Hagellach37, line 70: again here: do we want the splashscreen or not?

Hagellach37:
Hey @laurentS ,
most things are clear. Just some questions above on node version and if we want to have the splashscreen or not.

.flowconfig
laurentS: Honestly, I'm not entirely sure, I just followed the upgrade docs. flow was not on a priority list as it's never blocking releases anyway. I've put a lot of `$FlowFixMe` comments in the code just to get CI to pass, I suggest we leave resolving these for a separate PR, once we're back on track.

.github/workflows/android.yml
laurentS: Indeed, I think we can upgrade again, the downgrade was just to test if this was causing crashes. Done in 3b5bf5bd

.github/workflows/ios.yml
laurentS: Same answer as above :)

.github/workflows/ios.yml
laurentS: yes, done!

.github/workflows/ios.yml
laurentS: Removed, good catch.

android/app/src/main/AndroidManifest.xml
laurentS: Yes, this boils down to RNfirebase having to be upgraded to v6, in which notification management has been extracted to a separate component. I will have to put them back in, for the third time :(

android/app/src/main/java/org/missingmaps/mapswipe/MainActivity.java
laurentS: Two different things at play here:
- RN's import system works slightly differently now, so this line is not needed
- the splashscreen package we were using is not supported anymore, so I've removed it (first to check if it was causing the crash, but also to replace it with a properly maintained one, which I'll do asap)

android/app/src/main/java/org/missingmaps/mapswipe/MainApplication.java
laurentS: See point above: the import system is slightly different in this version of RN, we still very much want sentry to report errors, as we'd be virtually blind without it.

android/app/src/main/java/org/missingmaps/mapswipe/MainApplication.java
laurentS: Yes, see previous comments, except for notifications, everything else is still there, but these lines are not needed anymore.

ios/Podfile
laurentS: Flipper is RN's new debugger (for javascript, I think). I've not even had a chance to look at it yet, but apparently it's good.

src/shared/views/Welcome.js
laurentS: It's coming back very soon, but I think I'll open a separate PR for it, just so we can merge this and get back to normal life.

laurentS:
Comments addressed, and changes pushed back, if CI passes and you have no other comments, I think we're good to merge.

Hagellach37:
Hey,
yep let's merge this one.

Hagellach37 (approved):

PR #528, opened 2021-07-07
Fix Splashscreen

### LINKED ISSUE #526: Setup splashscreen with bootsplash
As part of upgrading to RN 0.64.2, I removed the splashscreen, as it was outdated and not maintained for the past 2 years or so.
Replacing the dependency with https://github.com/zoontek/react-native-bootsplash/ should get things back on track again.

laurentS:
This PR fixes #526 
Tested on both android device and ios (simulator only).

Please check that the splash screen logo size is acceptable on your device (it should be about the same size as before).


Hagellach37 (approved):
Hey @laurentS ,
looks all good to me.

I've checked on iOS and Android and both work as expected. Nice to have the splashscreen back. :)

Issue #522, opened 2021-07-01
user profile text not displayed


Hagellach37:
**Describe the bug**
the text on the user profile is not displayed properly

**App version**
android  2.0.11(24)-beta

**To Reproduce**
Please provide steps to reproduce the problem:
1. open the app
2. swipe to the user profile

**Screenshots**
![image](https://user-images.githubusercontent.com/7045979/124120331-f1ef0600-da73-11eb-8379-4375e16c2abf.png)

**Smartphone (please complete the following information):**
 - samsung galaxy tab

**Additional context**
- only happens in android not ios

Issue #86, opened 2019-07-11
Move away from react-native-scrollable-tab-view


laurentS:
This is another "Lean Core" issue (or related). We use https://github.com/ptomasroos/react-native-scrollable-tab-view for the main menu (to do the left/right scrolling between the list of projects and the "more" page). This relies on the `viewPagerAndroid` component (https://github.com/react-native-community/react-native-viewpager) which has been extracted from the core RN repo. This causes a deprecation warning in RN 0.59, and a crash in 0.60.

The deprecation warning is fixed for now, but this component is not maintained, and will cause issues sooner or later.

laurentS:
Development seems to have resumed on the component, so we probably only need to upgrade to the latest version.

PR #535, opened 2021-07-15
Fix navigation and display bugs

### LINKED ISSUE #523: swiping to user profile triggers tap
**Describe the bug**
when swiping to the user profile, a project page is opened.

**App version**
android 2.0.11(24)-beta

**To Reproduce**
Please provide steps to reproduce the problem:
1. open the app.
2. swipe to the user profile. (e.g. swipe in the middle of the screen where there is a project)
3. the project page opens, but not the user profile

**Screenshots**

https://user-images.githubusercontent.com/7045979/124120941-9bce9280-da74-11eb-81a9-cc4f62a6ef16.mp4

**Smartphone (please complete the following information):**
- samsung galaxy tab

**Additional context**
- not a problem on ios

laurentS:
Fixes #523 and #522, as well as #86 by integrating the code of this unmaintained component into our code base.

This is probably a little brutal, but the unmaintained [react-native-scrollable-tab-view](https://github.com/ptomasroos/react-native-scrollable-tab-view) relies on another component (pager-view) which we already use for the app anyway, and is itself upgraded regularly and used by a bigger community.
We can simplify the code brought in by removing all the options and parts we don't use, or just update everything and republish ;)


.github/workflows/ios.yml
```diff
@@ -33,6 +33,15 @@ jobs:
       - name: Install cocoapods
         run: gem install xcpretty cocoapods
 
+      - name: TEMP manually run pod install
```
Hagellach37, line 36: Did you add this again on purpose?

laurentS: Yes, it's not ideal, but it's the only way I've found to get an updated Podfile.lock without using a mac.

PR #538, opened 2021-07-16
fix import of Modal component

Hagellach37:
@laurentS can you check if this is enough?


laurentS (approved):
I can't see any other `require` calls, so I guess it's fine?

PR #554, opened 2021-07-21
don't display continue mapping button

Hagellach37:
Do not display the "continue mapping" button at the end of a mapping session, when the user has already mapped all groups of the 15 groups that are available. In that situation the user can only use the "Complete Session" button and will be navigated back to the projectView screen.

image


Hagellach37:
I do not understand why the Android build failed.

src/shared/views/ChangeDetection/TaskList.js
```diff
@@ -40,6 +40,7 @@ type Props = {
     closeTilePopup: () => void,
     openTilePopup: () => void,
     zoomLevel: number,
+    continueMapping: boolean,
```
laurentS, line 43: Could you maybe rename this to `canContinueMapping`? I think it makes it clearer what the variable means, no?

src/shared/common/firebaseFunctions.js
```diff
@@ -126,6 +131,10 @@ export const mapStateToPropsForGroups =
             const groupsToPickFrom = groupsAvailable.filter(
                 g => !groupsMapped.includes(g),
             );
+            if (groupsToPickFrom.length === 1) {
```
laurentS, line 134: Is there any possibility that the length of the list drops from 2 to 0 without ever being 1? I don't think so, but just double checking :)

src/shared/common/firebaseFunctions.js
```diff
@@ -115,6 +115,11 @@ export const mapStateToPropsForGroups =
                     data[prefix][projectId]);
             }
         }
+
+        // Here we define a boolean value 'continueMapping' which will tell us,
+        // if there are more groups available for mapping for this user.
+        let continueMapping;
+        continueMapping = true;
```
laurentS, line 122: Any reason not to write `let continueMapping = true`?

laurentS (changes requested):
Looks cool, I like the idea! Just a couple of minor comments around naming things to avoid confusion.

src/shared/common/firebaseFunctions.js
Hagellach37: no particular reason. so I will adjust as you recommended

src/shared/common/firebaseFunctions.js
Hagellach37: good catch. I will change to <= 1

src/shared/views/ChangeDetection/TaskList.js
Hagellach37: good idea. will rename this in all places where it occurs

Hagellach37:
Hey @laurentS ,

I've adjusted this based on your comments. I also added this for the building footprint project type. When you want to test this locally, it is handy if you do not get 15 groups but only 3 or so. You can adjust this in `firebaseFunctions.js`

Hagellach37:
closing this in favor of #571

PR #555, opened 2021-07-21
Fix loading screen while mapping

Hagellach37:
Here I change the place where we store the group information in the redux store. I do this to avoid the loading screen, that is triggered by an update of the project data in firebase, e.g. when the progress is updated.

I also added the logic to not download the tasks for change detection projects.

The reason for the loading screen error might have been caused by the following. As we had a nested data structure in the redux store, the project information and thus also the nested groups and tasks have been overwritten and thus were not available anymore.

We had a data structure like this:
* projects/projectId/groups/groupId/tasks

So when you set something for projects, which does not contain the information on groups and tasks, you will erase the information we had in there before.

* [x] check if this has side effects on the Footprint project type
* [x]  check flow


Hagellach37:
closing this in favor of #571

PR #556, opened 2021-07-23
Remove firebase tracking

laurentS:
This PR tries to unlink as much personal data as possible in firebase analytics.
With these changes, if I understand the docs correctly (which is a bit of a challenge in itself), it looks like firebase analytics will not be able to link personal identifiers with activity anymore. In practice, I think this means that analytics would report something like where the user is (country-level, based on anonymised IP address), and what action they took in the app ("start mapping" for instance)

### DIFF
--- a/docs/privacy.md
+++ b/docs/privacy.md
@@ -49,6 +49,7 @@ In short:
 - Apple's location service is not used, country is derived from an anonymized IP address
 - As we don't collect IDFA, we do not have access to age, gender, interests of the user ([source](https://support.google.com/firebase/answer/9268042?visit_id=637626397084989644-2548047988&rd=1))
 - We removed IDFV [collection](https://firebase.google.com/docs/analytics/configure-data-collection?platform=ios#disable_idfv_collection) in analytics. IDFV is an apple identifier that seems to uniquely identifier a device, and would allow [tracking usage of various apps of the same provider](https://developer.apple.com/forums/thread/659416) (for instance, track an iOS user across BRC's apps). We don't need this.
+- Analytics data is not used for [personalised advertising](https://firebase.google.com/docs/analytics/configure-data-collection?platform=ios#disable_personalized_advertising_features), although we never linked an ad account to the analytics account for MapSwipe.
 
 #### Android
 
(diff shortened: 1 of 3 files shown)


ios/mapswipe/Info.plist
```diff
@@ -39,6 +39,10 @@
 	</dict>
 	<key>NSLocationAlwaysUsageDescription</key>
 	<string>MapSwipe doesn't actually use your location.</string>
+	<key>GOOGLE_ANALYTICS_IDFV_COLLECTION_ENABLED</key>
+	<false/>
+	<key>GOOGLE_ANALYTICS_DEFAULT_ALLOW_AD_PERSONALIZATION_SIGNALS</key>
+	<false/>
```
Hagellach37, line 45: for android you disable three things, here only two. Is there a particular reason for this?

Is it because the collect different things on Android '(adid, ssaid) and iOS (idfv, ad personlization signals)?

docs/privacy.md
```diff
@@ -0,0 +1,61 @@
...
+- Apple's location service is not used, country is derived from an anonymized IP address
+- As we don't collect IDFA, we do not have access to age, gender, interests of the user ([source](https://support.google.com/firebase/answer/9268042?visit_id=637626397084989644-2548047988&rd=1))
+- We removed IDFV [collection](https://firebase.google.com/docs/analytics/configure-data-collection?platform=ios#disable_idfv_collection) in analytics. IDFV is an apple identifier that seems to uniquely identifier a device, and would allow [tracking usage of various apps of the same provider](https://developer.apple.com/forums/thread/659416) (for instance, track an iOS user across BRC's apps). We don't need this.
+- Analytics data is not used for [personalised advertising](https://firebase.google.com/docs/analytics/configure-data-collection?platform=ios#disable_personalized_advertising_features), although we never linked an ad account to the analytics account for MapSwipe.
```
jhenshall, line 53: Would remove 'although' as it implies we once did. Use full stop instead:
Analytics data is not used for [personalised advertising](https://firebase.google.com/docs/analytics/configure-data-collection?platform=ios#disable_personalized_advertising_features). We have never linked an ad account to the analytics account for MapSwipe.

ios/mapswipe/Info.plist
Hagellach37: we removed IDFA already for iOS

PR #557, opened 2021-07-23
Upgrade dependencies

laurentS:
Upgrade a bunch of libraries, this fixes no bugs, unfortunately :)


src/shared/common/BottomProgress.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import * as React from 'react';
-import * as Progress from 'react-native-progress';
+import ProgressBar from 'react-native-progress/Bar';
```
Hagellach37, line 3: Is there a good way to check for things like this? Or just trying the app and noticing that some errors will occur?

laurentS: check the release notes, that's the ideal way :)

Hagellach37:
Hey @laurentS ,
maybe a general question and more in idea than actual solution: In which situations would it make sense to go back to an earlier version of a library?

For instance we could try this for react-native-redux or another library related to the bug.

src/shared/common/BottomProgress.js
Hagellach37: okay good.

Hagellach37 (approved):

Issue #224, opened 2020-06-05
Imagery credits are missing in CD projects


laurentS:
**Describe the bug**
After replacing the help modal box with a separate screen, the credits for CD imagery are gone. They need to be added again in the help screen.

**App version**
Starting with `2.0.3(X)-beta`

PR #564, opened 2021-08-18
add imagery credits for change detection project type #224

Hagellach37:


Hagellach37:
Hey @laurentS,

the ios build failed and it says: `Your certificate 'MD6M8684YW.cer' is not valid, please check end date and renew it if necessary`.

What does this mean?

laurentS:
> the ios build failed and it says: `Your certificate 'MD6M8684YW.cer' is not valid, please check end date and renew it if necessary`.
> 
> What does this mean?

Moved this to #565

laurentS:
> this adds the imagery credits to the change detection project mapping view. I'm not sure what's the best placement of the credit label.
> 
> Maybe we could also think about moving this code into a new component?
> 
> What do you think @laurentS ?

Does the credit text have to be on this screen? I feel like it adds clutter. Could we have it on a separate info box, like with the main projects? Accessed via the "i" icon top right, maybe?

src/shared/views/ChangeDetection/index.js
```diff
@@ -50,6 +29,7 @@ type Props = {
 export default class ChangeDetectionScreen extends React.Component<Props> {
     randomSeed: number;
 
+    // TODO: this is not used at the moment.
```
Hagellach37, line 32: eventually we could also remove the code below as it is not used anymore

src/shared/views/ChangeDetection/Body.js
```diff
@@ -304,7 +306,10 @@ class _ChangeDetectionBody extends React.Component<Props, State> {
                 <View>
                     <TouchableWithoutFeedback
                         onPress={() => {
-                            navigation.push('CDInstructionsScreen');
+                            navigation.push('CDInstructionsScreen', {
```
laurentS, line 309: Can this be replaced with `onInfoPress` as the code looks exactly the same?

src/shared/views/ChangeDetection/InstructionsScreen.js
```diff
@@ -118,6 +121,13 @@ class CDInstructionScreen extends React.Component<Props> {
                 </View>
 
                 <ScrollView style={styles.container}>
+                    <Text style={styles.header}>
+                        <Trans i18nKey="CDInstructionScreen:imageryCredits">
```
laurentS, line 125: Any reason why you use `Trans` here instead of the `t()` function? The text is simple enough for the latter, I think.

src/shared/views/ChangeDetection/index.js
laurentS: Not exactly sure which code you're referring to, but if it's not used, I'd say we remove it now. Or are you planning on bringing it back soon?

laurentS (approved):
A couple of questions, but otherwise looks good to me :)

src/shared/views/ChangeDetection/index.js
Hagellach37: I think I will not add this back soon. So let's just remove it for now.

src/shared/views/ChangeDetection/InstructionsScreen.js
Hagellach37: I'm not sure if there is any difference between the two. For me trans was sometimes a bit more obvious as you can place default values. That helped during the initial implementation, but now I could also use only t().

Or is there any other difference between the two?

src/shared/views/ChangeDetection/Body.js
Hagellach37: good catch.

Hagellach37:
alright, thanks for the feedback. I adjusted the code based on your suggestions. :)

Hagellach37:
Hey @laurentS ,
I think that we've never merged this PR because we had some issues with the ios builds. But now these should be working and maybe you could give this one a last check.

Then we merge it into dev?

PR #572, opened 2021-08-21
Renew certs

laurentS:
DO NOT MERGE!!

This PR is simply there to keep a trace of how we renew ios certs, as this will happen every year around summer time.
Hopefully this branch is simple enough that the commits in it can be cherry picked, or it can be rebased or something, to do the process again next time :)


laurentS:
fastlane issue https://github.com/fastlane/fastlane/issues/19266 that details the problem we're hitting with pushing new certs back to gitlab.

Issue #562, opened 2021-07-31
Reduce app package size 


laurentS:
The apk weighs over 50MB, which is ludicrous.
For comparison, wikipedia is ~10, the github official client 8.

Not sure what the solution is, proguard might help.

laurentS:
Some ideas:
- analyze JS bundle content: https://create-react-app.dev/docs/analyzing-the-bundle-size/ or https://github.com/IjzerenHein/react-native-bundle-visualizer
- use Hermes as the JS engine: https://reactnative.dev/docs/hermes
- enable proguard: https://reactnative.dev/docs/signed-apk-android#enabling-proguard-to-reduce-the-size-of-the-apk-optional
- possibly use google's AAB packaging format to split the app per architecture (this seems to have a massive impact, but also means the APK is not the standard android one anymore) https://reactnative.dev/docs/signed-apk-android#generating-the-release-aab
- review assets, images mostly, which might need some recompressing

laurentS:
Using AAB requires letting google manage signing keys, which means we lose control over the process and cannot distribute a compatible APK outside the playstore. This would probably cause a lot of confusion among users as the app would appear to be a different one if downloaded directly or via playstore. There are probably ways to reduce the package size without going to such extremes.

laurentS:
Other ideas:
- drop unused architectures if usage stats suggest we can. Currently we build for `x86_64`, `x86`, `arm64_v8a` and `armeabi_v7a`. Each weighs between 6 and 8MB.
- switch to hermes JS engine, might save about 8MB
- the JS bundle weighs about 600kB, so we won't be able to gain much with it
- using proguard to remove a bunch of unused classes from DEX files. maybe a few 100kB again
- we might be able to gain a bit more on images by resizing/recompressing them

laurentS:
> @laurentS should we close this one? as you've merged this: #594

I'd leave it open, there are still a few things we can do to reduce the size of the apk in particular. Switching to Hermes would be very helpful, also for performance reasons, and it's apparently pretty stable by now.
I might give it a try at some point soon, as switching seems as simple as changing a bool to `true` in the code :)

PR #594, opened 2021-11-28
Resize level badges to 10% of original size

laurentS:
This PR works towards #562 by resizing the badge sizes which I found to weigh about 15MB using android studio's APK analyzer (docs [here](https://developer.android.com/studio/debug/apk-analyzer):
image

On the screenshot above, `res` contains all the images in the app. The 36 biggest ones are the level badges, which all weigh between 450 and 600kB because we have them saved as 2048x2048px images. We only display them as 150x150px in the app, so I resized to 205x205px (in case some phones have bigger screens than mine), which shaves about 15MB off the resulting APK (I don't know about the iOS IPA)

@jhenshall I'd love it if you could point your x-ray eyes at 2.1.6(3) and let me know if you can spot any difference (you shouldn't find any, hopefully).


Hagellach37:
Hey, looks good to me. You cancelled the tests, but I guess as we didn't change anything else in the code this should be fine.

Hagellach37 (approved):
Hey,
only one quick comment. You still keep the badges in the fullsize in the repo. They will not be part of the apk, right?

laurentS:
> Hey, only one quick comment. You still keep the badges in the fullsize in the repo. They will not be part of the apk, right?

I kept them in there so that we can create other sizes in the future from them. As far as I understand from checking with the APK inspector, the files don't end up in there, so it should be fine. (In my test, the APK size dropped from 55 to 40MB approx, which matches the drop in filesize.)

jhenshall (approved):
Nice one man! So I can see a small difference when comparing the screenshots of page but it is negligible for the saving in space. I wonder if bigger screen sizes might show them bigger though and so more noticeable? I think fine to press on but definitely keep the originals as you've said in case we need to resize in future.

PR #608, opened 2022-01-14
Prepare for release 2.1.8

laurentS:
This is essentially the same version as 2.1.7 (which was not actually published).

Proposed release notes:
> We have added support for a new project type which will help with reviewing OSM data quality.


Hagellach37 (approved):
Hey @laurentS ,
let's merge it. :)

I think there are a few things we need to work on in one of the next releases. I think mainly we need to think about how we can have the tutorial a bit more generic so that it fits also for other types of objects. 

Some things are easy, e.g. replacing the hard coded example image. We already have this in Firebase for the tutorial. Other things will be a bit more difficult such as translation.

I think ideally you could also create a dev release, so that we can test the new production version also in dev. I think the last dev release does not reflect the latest updates you did.

Issue #536, opened 2021-07-15
Get rid of apsl-react-native-button dependency


laurentS:
The [package](https://github.com/APSL/react-native-button) hasn't been updated in over 3 years, and is likely to break sooner or later. It's a pretty straightforward component which we could easily vendor in.

PR #616, opened 2022-02-01
Draft: Button replacement

ElJocho:
Solves: 
- #536

currently I only changed the buttons in a single file, so we can test it there; before merging this I'd change all the dependencies.

I only implemented features which are used in the app to keep the code lightweight, if we need any more features from the original dependency just mention it :).


Hagellach37:
one quick comment: before you commit you should run `yarn lint` and `yarn flow` to be able to pass the first check about Syntax and Linting

src/shared/common/Button.js
```diff
@@ -0,0 +1,68 @@
+import React, {Component} from 'react';
```
laurentS, line 1: Can you add a `// @flow` on the first line of the file to make sure flow looks at it?

src/shared/views/MoreOptions.js
```diff
@@ -29,6 +28,7 @@ import {
     COLOR_RED_OVERLAY,
     COLOR_WHITE,
 } from '../constants';
+import Button from '../common/Button';
```
laurentS, line 31: Does this change have any impact on the visual aspect of the button for you? should we expect any differences? One of the difficult parts with the app is that each phone seems to have a slightly different screen size, which leads to designs potentially looking a bit different on various models.
RN is supposed to help with this, but I've often hard coded numbers like sizes and stuff, only to find out that it didn't work on such and such phones.

src/shared/common/Button.js
```diff
@@ -0,0 +1,68 @@
...
+        android_ripple={{color:'light-gray'}}
+        {...extra_attributes}
+      >
+        {this._renderChildren()}
+      </Pressable>
+    )
+  };
+}
+
+const styles = StyleSheet.create({
```
laurentS, line 46: Can I suggest that we try to keep a consistent structure to component files?
I think for most of them, the stylesheet is at the top, maybe you can follow that same logic? It makes it much easier to navigate the codebase when we know where to find things :)
(there's no reason why top or bottom is better, it's purely standardization)

laurentS (changes requested):
As @Hagellach37 mentioned, make sure lint/flow pass, they pick up a ton of small things, like the fact that you used an indent of 2 instead of 4 :)

Is this code you've adapted from the original component? If so, it can be nice to mention it somewhere.
Then for the missing features, I think it can happen progressively, as they become necessary.

src/shared/views/MoreOptions.js
ElJocho: No, I was quite happy when nothing (obvious?) changed once I adjusted the import

src/shared/common/Button.js
```diff
@@ -0,0 +1,83 @@
...
+            if (typeof item === 'string' || typeof item === 'number') {
+                const element = (
+                    <Text style={[styles.text, textStyle]} key={item}>
+                        {item}
+                    </Text>
+                );
+                childElements.push(element);
+            } else if (React.isValidElement(item)) {
+                childElements.push(item);
+            }
```
laurentS, line 53: Would it be worth throwing an error message here if `item` is not valid?
```suggestion
            } else {
                console.log("item passed to Button is not valid.");
            }
```

src/shared/common/Button.js
```diff
@@ -0,0 +1,83 @@
...
+        });
+        return childElements;
+    };
+
+    render(): React.Node {
+        const { onPress, style, textStyle, testID, isDisabled, children } =
+            this.props;
+        if (isDisabled === true) {
```
laurentS, line 61:
```suggestion
        if (isDisabled) {
```
if it's a boolean, you can write this instead, a bit lighter on the eye :)

src/shared/common/Button.js
```diff
@@ -0,0 +1,83 @@
...
+type Props = {
+    children: any,
+    onPress: () => any,
+    textStyle: Object,
+    style: Object,
+    testID?: string,
+    isDisabled?: boolean,
+};
+
+class Button extends Component<Props> {
```
laurentS, line 40: If this is going to be a reusable component, maybe add a short docstring explaining the props that can be set on the button? Or maybe just add comments in the `type Props` definition above?

src/shared/common/Button.js
```diff
@@ -0,0 +1,83 @@
...
+    opacity: {
+        opacity: 0.5,
+    },
+});
+
+type Props = {
+    children: any,
+    onPress: () => any,
+    textStyle: Object,
+    style: Object,
```
laurentS, line 35: There are stricter types for this, like in https://github.com/mapswipe/mapswipe/blob/8d0f2f0ab64bf45dbbf7d69b8a74807107a90c68/src/shared/views/ChangeDetection/Tile.js#L13
or
https://github.com/mapswipe/mapswipe/blob/8d0f2f0ab64bf45dbbf7d69b8a74807107a90c68/src/shared/common/SatImage.js#L3

src/shared/common/Button.js
```diff
@@ -0,0 +1,83 @@
...
+        color: 'white',
+    },
+    opacity: {
+        opacity: 0.5,
+    },
+});
+
+type Props = {
+    children: any,
+    onPress: () => any,
```
laurentS, line 33:
```suggestion
    onPress: () => void,
```
I think this is the correct type. But you might want to double check elsewhere in the codebase, or in the docs https://flow.org/en/docs/react/types/

laurentS (approved):
I've added some annoying comments. I think we can merge after this :)

ElJocho:
@laurentS would it make sense to completly remove the dependency on this pr or should we wait for a later time?

laurentS:
> @laurentS would it make sense to completly remove the dependency on this pr or should we wait for a later time?

If you feel like doing it, I can't see any reason not to.

src/shared/common/Button.js
```diff
@@ -0,0 +1,83 @@
...
+            return (
+                <View style={[styles.button, style, styles.opacity]}>
+                    {this.renderChildren(textStyle, children)}
+                </View>
+            );
+        }
+        const extraAttributes = { testID };
+
+        return (
+            <Pressable
```
laurentS, line 71: I missed this the first time, but maybe it's worth looking at the built-in `Button` that RN offers? https://reactnative.dev/docs/0.65/button

ElJocho: Since you mentioned it in the first meeting on this issue I checked if this is an alternative, but the RN button does not seem to offer the android ripple effect as well as only the option to change the colors, but not anything else style wise.

[RN](https://reactnative.dev/docs/button) on this issue:
 ```
" A basic button component that should render nicely on any platform. Supports a minimal level of customization.

If this button doesn't look right for your app, you can build your own button using TouchableOpacity
or TouchableWithoutFeedback]."
```

File: src/shared/common/Button.js (hunk 7 above)
Comment (ElJocho) [reply src/shared/common/Button.js]: good catch, changed it to ViewStyleProp for Button and TextStyleProp for inner Button text

File: package.json [hunk 10]
```diff
@@ -32,7 +32,6 @@
     "@react-native-firebase/database": "^12.3.0",
     "@react-native-firebase/storage": "^12.3.0",
     "@sentry/react-native": "^3.1.1",
-    "apsl-react-native-button": "^3.1.1",
```
laurentS, line 35: Can you update the `yarn.lock` file and commit it?

laurentS (approved):
Lockfile is missing, but otherwise looks good to go!

ElJocho:
@Hagellach37 or any other ios user; can you check if the buttons look the same now as they did before I changed the dependency? Because I used a andoid_ripple=True on the Pressable component and I have no idea what that does to ios devices..

laurentS:
> Hey @ElJocho,
> 
> I'm happy to take a look. However for this you need to create a dev-release so that I can test it in testflight. If you don't know how to deploy one, you might ask @laurentS.
> 
> @laurentS do we have this workflow somewhere in the docs?

[This](https://github.com/mapswipe/mapswipe/blob/master/docs/devRelease.md) seems to be all we have. I was just chatting with @ElJocho about this :)

Issue #35, opened 2018-12-11
Use OSM to login to mapswipe app


laurentS:
Currently MapSwipe requires creating login and password. Missing Maps project is tightly coupled with OpenStreetMap project so it would easier for OSM users if MapSwipe could use OSM login.

(moved from https://github.com/mapswipe/app-feedback/issues/30)

laurentS:
Comment from @harry-wood on the other repo:

> I agree it would be good to offer OSM login as an alternative (dinky little button).
> 
> A lot of OpenStreetMap mappers might like to use that. It's worth noting that the real power of MapSwipe is to get people on board who are not already OpenStreetMappers (often because editing OpenStreetMap is too complicated). Even so I think this is a sensible thing to have as an option.

laurentS:
Docs: https://wiki.openstreetmap.org/wiki/OAuth
code example: https://github.com/osmlab/osm-auth

forteller:
I agree, and I was uncertain when I opened the app for the first time, whether I should log in with my OSM credentials or not. Before this is implemented, maybe there should be a short text informing you that you can not log in with your OSM account on the login screen?

Hagellach37:
adding this link here which might become useful when we want to link an OSM account to an existing MapSwipe account: https://firebase.google.com/docs/auth/web/account-linking

laurentS:
After discussion with @Hagellach37 today, here are some findings:

[This page](https://github.com/firebase/functions-samples/blob/main/instagram-auth/functions/index.js) shows an example of OAuth login using instagram. Login in with OSM should be fairly similar. It looks like this code should be able to run in a firebase function.

Some auth data would need to be stored in the RealtimeDatabase on firebase (OSM token, OSM user id, firebase user id...) so care must be taken to secure this correctly (permissions...).

Once a user is logged in to their OSM account, we can link this back to a MapSwipe account into which they are already logged in. The way to do this seems to rely on adding extra info to their existing MapSwipe account (OSM user id, OSM auth token), rather then the other way around. Mainly this is dictated by the need to preserve a user's history of contributions in MapSwipe (changing their user id in every contribution would be a ridiculous amount of work, if even possible).

Linking a MapSwipe account with an OSM account does not seem to require both having the same email, as we thought at one point. If the user is able to login to both, then we can link these 2 accounts.

So we would end up with 3 types of accounts:
- `MapSwipe only`: the user only has a MapSwipe account (or has not linked to OSM)
[...]

PR #618, opened 2022-02-10
Setup login with OSM

laurentS:
This PR addresses #35 and should be read in conjunction with https://github.com/mapswipe/python-mapswipe-workers/pull/521

This adds the ability to signup and login using an OSM account. For now, we deliberately use dev OSM accounts until we are comfortable that the workflow in the app is functioning properly.

Outstanding items:
- [x] configure deep link prefix based on the build type (prod/dev)
- [x] improve the workflow visually to give the user some sense of comfort that they have not landed on a scam site
- [x] Update the privacy policy to match the new situation. Tracked in #632 
- [ ] Store the fact that the user has agreed to our privacy policy on their phone so they don't need to check the box again? -> skipping this for now, as users are likely to only login very infrequently
- [x] Secure the various credentials used for this

Adding account linking for existing email/password accounts will be done in a separate PR once this is working reliably.

For reviewers, please check the documentation in particular for anything unclear. OAuth is a bit of a headache, and I'm not even sure I've implemented things correctly. I've read a variety of documents around this, and tried to explain what happens as clearly as possible in the docs. Please ask if you have any doubt.


src/shared/constants.js
```diff
@@ -60,3 +60,7 @@ export const tutorialModes = {
     // hint is shown if they press the "show answers" button
     hint: 'hint',
 };
+
+// the prefix used for our deeplink urls,
+// such as 'mapswipe://login/xxx'
+export const deeplinkPrefix = 'devmapswipe://';
```
Hagellach37, line 66: this should be `mapswipe://` for the production app, correct? how do we make sure we load the correct prefix?

Hagellach37 (commented):
Hey @laurentS ,
thanks for implementing this cool feature. :)

Most things are looking good to me. I've tested the workflow in the app and the only thing that confuses me is the fact that I get redirected to the app twice. It feels that it should be enough to do this only the first time, but maybe I'm wrong.

could you check if this is what is intended:
- tap agree to privacy notice --> then tap loginSignupWithOSM
- redirect to osm login page in safari --> add email and password and login to osm
- redirects to mapswipe app --> immediately redirects to safari and asks again to open in mapswipe again
- redirecty to mapswipe app project page

Other than that we need to add the translation strings but this should be a quick one.

Let me know if you have further questions or if my explanation above is not clear.

src/shared/constants.js
laurentS: Yes, I was thinking of using a separate prefix so that we can continue using both dev and prod apps side by side on a single phone. How we get this into the code is still a bit unclear to me. Here is easy, since it's a runtime evaluation. But the same value needs to be hardcoded at build time in `Info.plist` and `AndroidManifest.xml`. Maybe we can do something similar to how we set the values in `debugInfo.json` at build time. We could set `devmapswipe` as the default value, so that developers don't need to worry about it, and at build time on github actions, if it's a prod build, we replace the value with `mapswipe` in the 3 places we need.

src/shared/views/Login.js
```diff
@@ -107,6 +108,7 @@ const styles = StyleSheet.create({
 const SCREEN_SIGNUP = 0;
 const SCREEN_LOGIN = 1;
 const SCREEN_FORGOT_PASSWORD = 2;
+const SCREEN_OSM_LOGIN = 3;
```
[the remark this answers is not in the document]
laurentS: Yes, we could make this an enum but javascript does not actually have that :(

src/shared/views/Login.js
```diff
@@ -310,6 +323,58 @@ class _Login extends React.Component<Props, State> {
...
+        osmError: string,
+    ) => {
+        const { firebase, t } = this.props;
+        console.log('OSM callback', osmCode, osmState, fbToken, osmError);
+        // call /token now, which will get a token from firebase
+        // and redirect here via deeplink
+        if (fbToken === undefined) {
+            console.log('Fetching a token');
+            Linking.openURL(
+                `https://dev-mapswipe.web.app/token?code=${osmCode}&state=${osmState}`,
```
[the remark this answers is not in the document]
laurentS: Yes, same logic as your other comment about the prefix. It could be nice to deploy these functions under `mapswipe.org` or maybe a subdomain like `auth.mapswipe.org`, just to make it look a bit cleaner. I don't kow how much control we have over the domain though.

.github/workflows/ios.yml
```diff
@@ -52,6 +52,13 @@ jobs:
           GIT_TAG: ${{ steps.vars.outputs.tag }}
         run: sed -i '' "s/COMMIT/$GITHUB_SHA/" debugInfo.json; sed -i '' "s/TAG/$GIT_TAG/" debugInfo.json
 
+      - name: Hardcode OAuth prefix and urls for production version
+        if: startsWith(github.ref, 'refs/tags/') && !endsWith(github.ref, '-beta')
+        run: |
+          sed -i '' "s/string>devmapswipe/string>mapswipe/" ios/mapswipe/Info.plist
+          sed -d '' "s/dev-mapswipe.web.app/mapswipe.web.app/" debugInfo.json
```
Hagellach37, line 59: I'm not sure if `mapswipe.web.app` is the right url for production. It could also be `msf-mapswipe`. However, the url is already in use by the leaderboard... https://msf-mapswipe.web.app/

src/shared/views/Login.js
```diff
@@ -310,6 +323,58 @@ class _Login extends React.Component<Props, State> {
...
+            );
+        } else if (fbToken !== undefined) {
+            firebase
+                .auth()
+                .signInWithCustomToken(fbToken)
+                .then(userCredentials => {
+                    // Signed in
+                    const username = userCredentials.user.user.displayName;
+                    MessageBarManager.showAlert({
```
Hagellach37, line 349: for some reason this is not displayed for me.

.github/workflows/ios.yml
laurentS: Good question. I didn't think about it. It looks like we could add `mapswipe.web.app`, or we could do `auth.mapswipe.org` which would look more legit for users who pay attention.

laurentS: Let's use `dev-auth.mapswipe.org` and `auth.mapswipe.org` for the dev and production apps respectively.

.github/workflows/android.yml
```diff
@@ -70,10 +70,17 @@ jobs:
         run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
 
       - name: Save git version so the app can show it to users
-        env:
-          # we need this hack because github default env vars don't include the tag itself
-          GIT_TAG: ${{ steps.vars.outputs.tag }}
```
Hagellach37, line 75: why did you remove these lines? Don't we need the GIT_TAG anymore? It is still used in the run command below. You also did not remove it in the ios.yml file.

.github/workflows/ios.yml
```diff
@@ -128,7 +135,7 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/')
         uses: softprops/action-gh-release@v1
         with:
-          prerelease: endsWith(github.ref, '-beta')
+          # prerelease: endsWith(github.ref, '-beta')
```
Hagellach37, line 138: why did you put this line into a comment?

locales/en.json
```diff
@@ -33,6 +33,8 @@
     "signUp": "Sign up",
     "loginExistingAccount": "Log in to an existing account",
     "login": "Login",
+    "loginSignupWithOSM": "Login with OpenStreetMap",
+    "OSMsignupExplanation": "You can signup to Mapswipe using your OpenStreetMap account. If you already have a MapSwipe account, this will not link them together! (this will come soon in a future release)",
```
Hagellach37, line 37: here it should be "can signup to MapSwipe" (capital S)

Hagellach37 (approved):
can you check the minor comments. Mainly the one about the android workflow in github actions.

.github/workflows/android.yml
laurentS: I don't know what happened here, but it should not have gone. Thanks for picking it up!

.github/workflows/ios.yml
laurentS: This was a test to check the `pre-release` flag on https://github.com/mapswipe/mapswipe/releases. I noticed that the android build was setting it correctly, but then the ios build (which always takes much longer) was switching it incorrectly, this was to check that my impression was right (it is). Now I need to figure out a fix :)

laurentS: should be ok now.

Issue #619, opened 2022-03-03
Wrong tasks in group results


Hagellach37:
**Describe the bug**
it seems that the app writes a results to a group which the task doesn't belong to. The last 4 tasks should belong to another group.

```
{
  "endTime" : "2022-03-02T05:39:06.870Z",
  "results" : {
    "t16798" : 1,
    "t16799" : 1,
    "t16800" : 1,
    "t16801" : 1,
    "t16802" : 1,
    "t16803" : 1,
    "t16804" : 1,
    "t16805" : 1,
    "t16806" : 1,
    "t16807" : 1,
    "t16808" : 1,
    "t16809" : 1,
    "t16810" : 1,
    "t16811" : 1,
    "t16812" : 1,
    "t16813" : 1,
    "t16814" : 1,
    "t16815" : 1,
    "t16816" : 1,
    "t16817" : 1,
    "t16818" : 1,
    "t16819" : 1,
    "t16820" : 1,
    "t16821" : 1,
    "t16822" : 1,
    "t16823" : 1,
    "t16824" : 1,
    "t16825" : 1,
    "t16826" : 1,
    "t16827" : 1,
    "t16828" : 1,
    "t16829" : 1,
    "t16830" : 1,
    "t16831" : 1,
    "t16832" : 1,
    "t16833" : 1,
    "t16834" : 1,
    "t17427" : 1,
    "t17428" : 1,
    "t17429" : 1,
    "t17430" : 1
  },
  "startTime" : "2022-03-02T05:38:57.185Z"
}
```

**App version**
not really clear, but probably 2.1.8

**Time of crash**
2022-03-02T05:38:57.185Z

**To Reproduce**
can't reproduce

userID: `ej8YSzSA4yS2WOPpTebnVP02GDX2`

**Additional context**
I've noticed this because it triggered an error in the backend workers when transfering the results to postgres.

```
DETAIL:  Key (project_id, group_id, task_id)=(-Mx0_479a3w9FD9oMv5o, g554, t17427) is not present in table "tasks".
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/mapswipe_workers/firebase_to_postgres/transfer_results.py", line 93, in transfer_results_for_project
    save_results_to_postgres(results_file)
  File "/usr/local/lib/python3.8/dist-packages/mapswipe_workers/firebase_to_postgres/transfer_results.py", line 292, in save_results_to_postgres
    p_con.query(query_insert_results)
[...]

PR #626, opened 2022-04-13
Draft: Alert wrong tasks no

ElJocho:
This is a proposal for how we could check for additional task and alert sentry. 
The question we asked ourselves was, how we would test if it works if we can not reproduce the issue.

Addresses:
- #619


laurentS (approved):
So for now, you only log an error, but you don't prevent the upload from happening. Are you also changing the backend so that the invalid group is cleaned or ignored or something, so that it doesn't crash?

ElJocho:
Yes, handling in backend is addressed with this pr: https://github.com/mapswipe/python-mapswipe-workers/pull/530

PR #627, opened 2022-04-15
remove outdated step from readme overview

danbjoseph:


laurentS (approved):
Thanks for the edits!

PR #633, opened 2022-05-16
Prepare for release 2.2.0

laurentS:
This release brings the long awaited ability to login to MapSwipe using an OpenStreetMap account!
Linking existing accounts is not possible yet, but should come in the next release.

It also includes a number of security fixes, as well as updated translations for Czech, Dutch and Hungarian.

For project managers, it is now possible to force users through the tutorial (if they haven't taken it before) for all project types.

And there's a new `Button` component which frees us from one unmaintained dependency.


.github/workflows/ios.yml
```diff
@@ -9,7 +9,7 @@ on: [push]
 jobs:
   release-ios:
     name: Build and release iOS app
-    runs-on: macOS-latest
+    runs-on: macos-latest
```
Hagellach37, line 12: interesting that this changed

Hagellach37 (approved):
looks good! let's merge it

Issue #656, opened 2022-11-30
Add a validation check with app/playstore prior to publishing a release


laurentS:
Our current process to publish a new release looks like:
- open PR from `dev` to `master`
- review/merge
- add a new version commit+tag on `master` via our release script
- wait for CI to build and publish to both stores
- be sad when either of them (or both) fail :sob: 

At this point, we've waited several hours, and our release hasn't gone through. In some cases, like today, one store accepts the release, but not the other one. To keep release numbers in sync, we now have to figure out what the problem is, fix in `dev` and restart the whole process.

It would help to have a way to validate that a release will be accepted by the stores *before* we send it "officially", which would save hours. I suspect there might be a way to do this with `fastlane`, but haven't found the solution yet.

laurentS:
Looks like an [option for this](https://github.com/fastlane/fastlane/pull/20247) was added recently to `fastlane` for apple.
`precheck` would probably be helpful as well (iOS again) http://docs.fastlane.tools/actions/precheck/

laurentS:
and maybe `validate_only` on https://docs.fastlane.tools/actions/upload_to_play_store/ for android?

laurentS:
This page might be useful to check manually as part of our release process https://developer.android.com/distribute/play-policies

laurentS:
following the mixup with #659 , I tried running the fastlane `deploy` lane we normally use to publish to the playstore, but setting `validate_only=true` locally.

This is the output from fastlane on my machine, after building the apk locally using `yarn buildProd`. My understanding is that google is happy.
We should probably run something like this 
```
[12:18:41] ❯ bundle exec fastlane deploy
[✔] 🚀
[12:32:38]: ------------------------------
[12:32:38]: --- Step: default_platform ---
[12:32:38]: ------------------------------
[12:32:38]: Driving the lane 'android deploy' 🚀
[12:32:38]: ----------------------------------
[12:32:38]: --- Step: upload_to_play_store ---
[12:32:38]: ----------------------------------

+------------------------------------+---------------------------------------------------------------------+
|                                        Summary for supply 2.211.0                                        |
+------------------------------------+---------------------------------------------------------------------+
| track                              | internal                                                            |
| apk                                | app/build/outputs/apk/production/release/app-production-release.apk |
| validate_only                      | true                                                                |
| package_name                       | org.missingmaps.mapswipe                                            |
| release_status                     | completed                                                           |
| metadata_path                      | ./fastlane/metadata/android                                         |
[...]

PR #660, opened 2022-12-06
Prepare for release 2.2.4 - Fix broken build on android

laurentS:
This is a bugfix PR to release a new functional version for android.
There are no other visible changes than the fact that the app is working :)


laurentS:
I managed to run a `validate_only` fastlane job to confirm that google playstore would accept the build without screaming. See #656 for details on what I did.
As far as I understand it, this means the APK will be accepted when we push an actual release.

jhenshall (approved):
Nice one @laurentS for getting this fixed so quick! 👏 
An (obvious in hindsight) note to selves: we need to ensure regression testing before all prod releases. 🙂

PR #666, opened 2023-01-10
Fix problem with calendar heatmap

tnagorra:
Address
- https://github.com/mapswipe/mapswipe/issues/765
- https://github.com/mapswipe/mapswipe/issues/764

Changes
- Share algorithm used to create linechart in community dashboard to populate data for calendar heatmap
- Use custom Button in UserGroup component

### DIFF
--- a/src/shared/common/CalendarHeatmap.js
+++ b/src/shared/common/CalendarHeatmap.js
@@ -85,27 +85,20 @@ function getTimestamps(
 
     const timestamps = [sanitizedStartDate.getTime()];
 
-    let increment = 1;
-    while (true) {
+    const increment = 1;
+    while (sanitizedStartDate < sanitizedEndDate) {
         const myDate = new Date(sanitizedStartDate);
+
         if (resolution === 'year') {
-            myDate.setFullYear(sanitizedStartDate.getFullYear() + increment);
+            myDate.setFullYear(myDate.getFullYear() + increment);
         } else if (resolution === 'month') {
-            myDate.setMonth(sanitizedStartDate.getMonth() + increment);
+            myDate.setMonth(myDate.getMonth() + increment);
         } else {
-            myDate.setDate(sanitizedStartDate.getDate() + increment);
+            myDate.setDate(myDate.getDate() + increment);
         }
-        myDate.setHours(0);
-        myDate.setMinutes(0);
-        myDate.setSeconds(0);
-        myDate.setMilliseconds(0);
-
-        if (myDate > sanitizedEndDate) {
-            break;
-        }
-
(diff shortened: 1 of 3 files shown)


frozenhelium (approved):

src/shared/common/CalendarHeatmap.js
```diff
@@ -42,13 +42,80 @@ const styles = StyleSheet.create({
...
+    endDate: string | number | Date,
+    resolution: 'day' | 'month' | 'year',
+) {
+    const sanitizedStartDate = resolveTime(startDate, resolution);
+    const sanitizedEndDate = resolveTime(endDate, resolution);
+
+    const timestamps = [sanitizedStartDate.getTime()];
+
+    let increment = 1;
+    while (true) {
```
laurentS, line 89: Any risk of an infinite loop here? I'm thinking if the start and end dates are very far apart, or if `endDate < startDate` then you'd never exit and the app would hang. Is it worth adding some checks to prevent that?

frozenhelium: @laurentS since startDate and endDate are always 35 days apart, there isn't any chance of infinite loop here.

danbjoseph:
@samshara, this is marked as awaiting your review. can you please either review or remove yourself from the list for awaiting review?

samshara (approved):
Looks good to me

src/shared/common/CalendarHeatmap.js
laurentS: > @laurentS since startDate and endDate are always 35 days apart, there isn't any chance of infinite loop here.

They always are, until they are not. Some other dev might call this function without realising this expectation (which isn't documented), and now you have a bug :)

samshara: @laurentS we have removed the possibility of an infinite loop here.

src/shared/common/CalendarHeatmap.js
```diff
@@ -63,87 +135,93 @@ function CalendarHeatmap(props: Props) {
     const itemWidth = Math.min((screenWidth - SPACING_MEDIUM * 2 - 1) / 7, 38);
 
     const now = new Date();
+    now.setHours(0);
+    now.setMinutes(0);
+    now.setSeconds(0);
+    now.setMilliseconds(0);
+
```
bewakes, line 142: Since this is repeated most of the time. Let's use the golden rule of reusability and create a function for this. This just looks ugly appearing at lots of places.

bewakes (approved):

frozenhelium (approved):

PR #736, opened 2023-01-27
[WIP] upgrade of react-native to v0.70.6

laurentS:
This PR is the result of:
- running `npx react-native upgrade --version 0.70.6` which upgraded files partially.
- Using [https://react-native-community.github.io/upgrade-helper/?from=0.65.3&to=0.70.6](rn's upgrade-helper) to work out the missing bits and apply them manually (some iOS related parts have been omitted for now)
- manually working out a few fixes to get the code to compile using `yarn runAndroid`
- from there, run `yarn start` and fix errors as they come up. At this point, they seem to relate to outdated dependencies that I had to upgrade as I found them causing problems.

This is the latest error message I'm getting from `yarn start` with the code as committed:
```
info Reloading app...
 BUNDLE  ./index.js

 ERROR  Invariant Violation: ViewPropTypes has been removed from React Native. Migrate to ViewPropTypes exported from 'deprecated-react-native-prop-types'., js engine: hermes
 ERROR  Invariant Violation: Failed to call into JavaScript module method AppRegistry.runApplication(). Module has not been registered as callable. Registered callable JavaScript modules (n = 10): Systrace, JSTimers, HeapCapture, SamplingProfiler, RCTLog, RCTDeviceEventEmitter, RCTNativeAppEventEmitter, GlobalPerformanceLogger, JSDevSupportModule, HMRClient.
        A frequent cause of the error is that the application entry file path is incorrect. This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native., js engine: hermes
 ERROR  Invariant Violation: Failed to call into JavaScript module method AppRegistry.runApplication(). Module has not been registered as callable. Registered callable JavaScript modules (n = 10): Systrace, JSTimers, HeapCapture, SamplingProfiler, RCTLog, RCTDeviceEventEmitter, RCTNativeAppEventEmitter, GlobalPerformanceLogger, JSDevSupportModule, HMRClient.
        A frequent cause of the error is that the application entry file path is incorrect. This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native., js engine: hermes
```

## TODO
Continue upgrading dependencies to solve errors that pop up. This particular one seems related to `ViewPropTypes` that some older packages are still importing from the deprecated place.

The `AppRegistry` error *seems* to be related to `react-navigation` v4 which is not maintained anymore, so it wasn't upgraded to match the latest RN version. There might be other dependencies involved as well.

Relates to
- #647 
- #734


tnagorra:
@laurentS did you check if the android build is successful?

laurentS:
> @laurentS did you check if the android build is successful?

`yarn runAndroid` is successful in the sense that it compiles, but then `yarn start` fails with a bunch of errors in the JS code of dependencies.
`yarn buildDev` fails with an error on a regular expression that I suspect is not handled by `hermes`. I suspect that's only a minor issue, but there are most probably others hiding behind this one.

> Also, I have not tried this yet but `npx @rnx-kit/align-deps --requirements react-native@[major.minor]` can be used to upgrade dependencies (although not all packages are supported) More details here: [microsoft.github.io/rnx-kit/docs/guides/dependency-management](https://microsoft.github.io/rnx-kit/docs/guides/dependency-management)

Interesting! I had not heard of this. I'd be curious to see how it performs on headaches like API changes across major releases and such :) Please report back if you try it.

danbjoseph:
this is replaced by the work in #751

PR #746, opened 2023-02-09
Feature/upgrade incremental 1

tnagorra:


locales/es.json
```diff
@@ -1,7 +1,8 @@
 {
   "mainHeader": {
     "missions": "Misiones",
-    "more": "Más"
```
laurentS, line 4: Can you remove this file from this PR? It has nothing to do here, I guess it's because the base branch is wrong (the PR should be against `dev`).

package.json
```diff
@@ -42,7 +42,7 @@
     "postinstall-postinstall": "^2.1.0",
     "react": "17.0.2",
     "react-i18next": "^11.5.0",
-    "react-native": "^0.65.1",
+    "react-native": "0.66.0",
```
laurentS, line 45: I think you can save some time by upgrading to the latest patch version of each step, which would require just 1 upgrade per minor version, instead of dozens of them.

laurentS (changes requested):
The base branch of the PR should be `dev` pretty much in all cases :)

package.json
tnagorra: It does make sense.
Surprisingly all the changes made to [ios/mapswipe.xcodeproj/project.pbxproj](https://github.com/mapswipe/mapswipe/pull/746/files#diff-8fd159f5cf53c9ced0c8a959f2a9dac2490e172c9f3fc3966c73f6edd5b8f061) in 0.66.0 was removed in 0.66.5
So, it should be a bit easier as well.

locales/es.json
tnagorra: I can rebase the branch with dev.

tnagorra: Looks like it is up-to-date with `dev`.
Maybe the diff was because the base branch to merge was `master`

laurentS (approved):
Great to see this passing all CI checks! I'm assuming you've tested the app locally and it runs fine?

package.json
laurentS: Yes, that `pbxproj` file is a disaster to update. Well done fixing it!

tnagorra:
The android app runs fine. There was a minor rendering difference on InfoCard so we had to tweak the style.

But, we have not tested the ios app locally.

laurentS:
Can I suggest you release a dev build for each of these PRs when they seem ready, so we can have a few people testing both android and iOS versions? If they find no problem, then we merge the corresponding PR and move on.

Issue #639, opened 2022-08-29
Prevent spaces and other illegal characters in usernames


laurentS:
It is currently possible to create an account with a username that includes leading or trailing spaces, such as `User1<space>`.
This can lead to confusion, and should be prevented.

maxmaster55:
> includes leading or trailing spaces 
> 
can the username contain spaces other than this?

mugolnik:
Hi,
I’m a student working on contributing to open-source projects for a college software engineering course.
I see this issue never got closed, would it be okay if I looked into helping with it?
Thanks!

PR #838, opened 2024-03-19
fix: add username validation with simple Regex

Ceikry:
- Addresses #639 
- Adds unit test to assert standard validation rules (letters, numbers and underscores. No special symbols or spaces.)
- Moves the email-in-username check to before the main validation check so the specialized message can still appear.

### DIFF
--- a/src/shared/utils.js
+++ b/src/shared/utils.js
@@ -258,3 +258,8 @@ export function getTimeSegments(
         lastState,
     );
 }
+
+export function isValidUsername(username: String): Boolean {
+    const re = new RegExp('^[a-zA-Z0-9_]+$', 'g');
+    return re.test(username);
+}
--- a/src/shared/views/ChangeUserName.js
+++ b/src/shared/views/ChangeUserName.js
@@ -18,6 +18,7 @@ import {
 import PageHeader from '../common/PageHeader';
 import Button from '../common/Button';
 import type { NavigationProp, TranslationFunction } from '../flow-types';
+import { isValidUsername } from '../utils';
 
 const styles = StyleSheet.create({
     changeUserNameScreen: {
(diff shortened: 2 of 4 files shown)


src/shared/views/ChangeUserName.js
```diff
@@ -87,7 +88,7 @@ function ChangeUserName(props: Props) {
     const [updatePending, setUpdatePending] = React.useState(false);
 
     const handleConfirmButtonClick = React.useCallback(() => {
-        if ((newUserName?.length ?? 0) >= MIN_USERNAME_LENGTH) {
+        if (isValidUsername(newUserName)) {
```
tnagorra, line 91: We should still check for character length. There should be a validation to check that username should have at least MIN_USERNAME_LENGTH characters.

src/shared/views/Login.js
```diff
@@ -199,25 +200,27 @@ class _Login extends React.Component<Props, State> {
         const { firebase, navigation, t } = this.props;
         const { email, password, username } = this.state;
         const parent = this;
-        if (username !== null && username.length < MIN_USERNAME_LENGTH) {
+
+        if (username !== null && username.indexOf('@') !== -1) {
```
tnagorra, line 204: We do not need to check for '@' as it is already covered by the regex characters validation.

src/shared/utils.js
tnagorra, line 1: We also need to add this validation on sign up.

tnagorra, line 1: We also need to check if existing users already have these invalid characters. If so, they will no longer be able to sign in.

tnagorra (changes requested):

kopitek8:
Hey @Ceikry. We have reviewed your pull request and there are some action you need to take. Please **resolve the comments**, **rebase** the working branch with the `main` branch and finally **send** an updated pull request. Please do let us know if you have any questons.

PR #851, opened 2024-06-20
Fix IOS build

kopitek8:


puranban:
Fixed in this [PR](https://github.com/mapswipe/mapswipe/pull/850)

PR #860, opened 2024-07-19
Feature/accessibility instruction

puranban:
### Changes

- Add accessibility modal feature instruction for the first time only

### NOTE
- User can dismiss the accessibility instructions modal from all types of project.

### DIFF
--- a/src/shared/common/AccessibilityInfoModal.js
+++ b/src/shared/common/AccessibilityInfoModal.js
@@ -1,0 +1,160 @@
+import React, { useCallback, useEffect, useRef } from 'react';
+import Modal from 'react-native-modalbox';
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import { Text, View, StyleSheet, Image } from 'react-native';
+import { Trans } from 'react-i18next';
+
+import { COLOR_DEEP_BLUE } from '../constants';
+import Button from './Button';
+import BadImageIcon from '../views/assets/bad_image_new_icon.png';
+import QuestionMarkIcon from '../views/assets/question_mark_new_icon.png';
+import TickIcon from '../views/assets/tick_new_icon.png';
+
+const styles = StyleSheet.create({
+    modal: {
+        height: 'auto',
+        width: 300,
+        backgroundColor: '#ffffff',
+        borderRadius: 2,
+        justifyContent: 'center',
+        padding: 20,
+    },
+    header: {
+        fontWeight: '700',
+        color: '#212121',
+        fontSize: 18,
+    },
+    row: {
+        flexDirection: 'row',
+        justifyContent: 'flex-start',
+        marginTop: 15,
+    },
+    text: {
+        fontSize: 13,
+        fontWeight: '600',
+        color: '#50acd4',
+        paddingLeft: 5,
+    },
+    buttonText: {
+        fontSize: 13,
+        color: '#ffffff',
+        fontWeight: '700',
+    },
+    image: {
+        height: 30,
+        width: 30,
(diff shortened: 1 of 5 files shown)


src/shared/common/AccessibilityInfoModal.js
```diff
@@ -0,0 +1,168 @@
...
+        getData();
+    }, []);
+
+    const closeModal = useCallback(() => {
+        try {
+            AsyncStorage.setItem('visitedRoute', 'true');
+            modalRef.current?.close();
+        } catch (e) {
+            console.log('error', e);
+        }
```
samshara, line 85: maybe we can use `asyc` function here.

src/shared/common/AccessibilityInfoModal.js
```diff
@@ -0,0 +1,168 @@
...
+
+import { COLOR_DEEP_BLUE } from '../constants';
+import Button from './Button';
+
+/* eslint-disable global-require */
+const GLOBAL = require('../Globals');
+
+const styles = StyleSheet.create({
+    modal: {
+        height: GLOBAL.SCREEN_HEIGHT < 500 ? GLOBAL.SCREEN_HEIGHT - 50 : 'auto',
```
tnagorra, line 15: What is the logic behind this? What happens if height is zero or negative?

src/shared/common/AccessibilityInfoModal.js
```diff
@@ -0,0 +1,168 @@
...
+    },
+});
+
+function AccessibilityInfoModal() {
+    const modalRef = useRef();
+
+    const getData = async () => {
+        try {
+            //     Check if user has visited this route before
+            //     For simplicity, using a boolean flag stored in AsyncStorage
```
tnagorra, line 63:
```suggestion
            // Check if user has visited this route before
            // For simplicity, using a boolean flag stored in AsyncStorage
```

src/shared/common/AccessibilityInfoModal.js
```diff
@@ -0,0 +1,168 @@
...
+                <Text style={styles.text}>
+                    <Trans i18nKey="AccessibilityInstruction:tickIconInfo">
+                        Tap twice to turn the tile yellow and show a question
+                        icon
+                    </Trans>
+                </Text>
+            </View>
+            <View style={styles.row}>
+                <Image
+                    source={require('../views/assets/bad_image_new_icon.png')}
```
tnagorra, line 134: Let's move these require at the top. Why are we not using import syntax?

src/shared/common/AccessibilityInfoModal.js
```diff
@@ -0,0 +1,168 @@
...
+            </View>
+            <View style={styles.buttonContainer}>
+                <Button
+                    style={styles.closeButton}
+                    onPress={closeModal}
+                    textStyle={{
+                        fontSize: 13,
+                        color: '#ffffff',
+                        fontWeight: '700',
+                    }}
```
tnagorra, line 160: Let's move this style to the styles

tnagorra (changes requested):
Let's fix all of the issues before we can deploy the changes

src/shared/common/AccessibilityInfoModal.js
puranban: I don't think we need set modal  height according to the device, we can simply use `height: 'auto'` in this case.

@tnagorra what do you think?

tnagorra: I don't have much idea about the usecase here.

tnagorra (approved):

PR #869, opened 2024-07-30
Improve condition for username creation

puranban:
### Change

- Improve validation condition for username creation
- Add same validation on change username from profile section

### Note:

- Now username cannot contain space and capital letter (New feature)
- Username cannot contain `@` also existing behavoir

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -7,10 +7,10 @@
   "signup": {
     "chooseUsername": "Choose your username",
     "passwordError": "Password must be at least 6 characters long",
-    "usernameError": "Username must be at least 4 characters long and cannot contain space and capital letter",
+    "usernameError": "Username must be at least 4 characters long and cannot contain space and uppercase",
     "userNameExistError": "Username already exist, Please choose another username",
     "errorOnSignup": "Error on sign up",
-    "usernameErrorMessage": "Your username must be 4 characters or more and cannot contain space and capital letter",
+    "usernameErrorMessage": "Your username must be 4 characters or more and cannot contain space and uppercase",
     "usernameNotEmail": "Your username can not be an email",
     "usernamePublic": "Your username will be publicly visible",
     "success": "Success",
--- a/locales/en.json
+++ b/locales/en.json
@@ -7,9 +7,9 @@
   "signup": {
     "chooseUsername": "Choose your username",
     "passwordError": "Password must be at least 6 characters long",
-    "usernameError": "Username must be at least 4 characters long",
+    "usernameError": "Username must be at least 4 characters long and cannot contain space and capital letter",
     "errorOnSignup": "Error on sign up",
-    "usernameErrorMessage": "Your username must be 4 characters or more",
+    "usernameErrorMessage": "Your username must be 4 characters or more and cannot contain space and capital letter",
     "usernameNotEmail": "Your username can not be an email",
     "usernamePublic": "Your username will be publicly visible",
     "success": "Success",
--- a/locales/en.json
+++ b/locales/en.json
@@ -8,6 +8,7 @@
     "chooseUsername": "Choose your username",
     "passwordError": "Password must be at least 6 characters long",
     "usernameError": "Username must be at least 4 characters long and cannot contain space and capital letter",
+    "userNameExistError": "Username already exist, Please choose another username",
     "errorOnSignup": "Error on sign up",
     "usernameErrorMessage": "Your username must be 4 characters or more and cannot contain space and capital letter",
     "usernameNotEmail": "Your username can not be an email",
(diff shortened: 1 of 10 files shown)


samshara (dismissed):

locales/en.json
```diff
@@ -7,9 +7,10 @@
   "signup": {
     "chooseUsername": "Choose your username",
     "passwordError": "Password must be at least 6 characters long",
-    "usernameError": "Username must be at least 4 characters long",
+    "usernameError": "Username must be at least 4 characters long and cannot contain space and capital letter",
```
tnagorra, line 10:
```suggestion
    "usernameError": "Username must be at least 4 characters long and cannot contain space and uppercase",
```

locales/en.json
```diff
@@ -7,9 +7,10 @@
...
+    "userNameExistError": "Username already exist, Please choose another username",
     "errorOnSignup": "Error on sign up",
-    "usernameErrorMessage": "Your username must be 4 characters or more",
+    "usernameErrorMessage": "Your username must be 4 characters or more and cannot contain space and capital letter",
```
tnagorra, line 13:
```suggestion
    "usernameErrorMessage": "Your username must be 4 characters or more and cannot contain space and upperbase",
```

src/shared/views/Login.js
```diff
@@ -140,6 +156,18 @@ type State = {
     signupPPChecked: boolean,
 };
 
+const checkInvalidUserName = name => {
+    if (name !== null && name.length < MIN_USERNAME_LENGTH) {
+        return true;
```
tnagorra, line 161: Let's rename this function to validateUserName

src/shared/views/Login.js
```diff
@@ -140,6 +156,18 @@ type State = {
...
+    }
+
+    if (name !== null && name.length >= MIN_USERNAME_LENGTH) {
+        const validationRegex = /^[^A-Z\s]+$/;
+        return !validationRegex.test(name);
```
tnagorra, line 166: Instead of regex validation let's transform the username and compare if it has changed.

tnagorra (approved):

src/shared/utils.js
```diff
@@ -258,3 +262,37 @@ export function getTimeSegments(
...
 }
+
+export function validateUserName(name) {
+    if (!name) return false;
+
+    if (name.length < MIN_USERNAME_LENGTH) {
+        return false;
+    }
+
+    if (name.length >= MIN_USERNAME_LENGTH) {
```
tnagorra, line 273: We don't need this check. This will always be true.

src/shared/views/ChangeUserName.js
```diff
@@ -80,29 +89,45 @@ type InjectedProps = {
...
-                    () => {
-                        setUpdatePending(false);
-                        console.error('failed to change update user profile');
-                    },
-                );
+
+    const handleConfirmButtonClick = React.useCallback(async () => {
+        const isValidUserName = validateUserName(newUserName);
+        if (!isValidUserName) {
+            setShowUserNameError(true);
```
tnagorra, line 102: Why not just use MessageBarManager.showAlert?

tnagorra: This is not consistent with the Login page.

puranban: to  show error for already existed user in firebase we are using `MessageBarManager.showAlert`.
for user character validation we are showing error message as helper text

tnagorra: This should at least be consistent with the other page.

src/shared/views/Login.js
```diff
@@ -218,9 +221,22 @@ class _Login extends React.Component<Props, State> {
...
+                title: t('signup:errorOnSignup'),
+                message: t('signup:userNameExistError'),
+                alertType: 'error',
+                shouldHideAfterDelay: false,
+            });
+            this.setState({ loadingNext: false });
+            return;
+        }
+
+        this.setState({ loadingNext: true });
```
tnagorra, line 239:
```suggestion
```

src/shared/views/Login.js
```diff
@@ -240,6 +256,7 @@ class _Login extends React.Component<Props, State> {
                     projectContributionCount: 0,
                     taskContributionCount: 0,
                     username,
+                    usernameKey: username,
```
tnagorra, line 259: Let's add a NOTE why usernameKey = username in this case.

tnagorra: The comment is not correct. Use this instead:
We can assign username to usernameKey because we have already validated the username earlier

tnagorra (approved):

samshara (approved):

Issue #760, opened 2023-04-12
rename default branch


danbjoseph:
at some point it would be good to rename the default branch to "main" or similar

PR #871, opened 2024-08-14
Fix/documentation rename

puranban:
### Addresses

- https://github.com/mapswipe/mapswipe/issues/760

### Changes

- Rename master to main on all documentation


ios/mapswipe.xcodeproj/project.pbxproj
tnagorra, line 1: Any reason we have changes to files in ios/* and package.json?

samshara (approved):

samshara:
Have we created the `main` branch and set it as the default branch? @shreeyash07

PR #921, opened 2025-07-15
Fix/ios next image validate issue

AdityaKhatri:


frozenhelium (approved):

src/shared/views/ImageValidation/IndividualTask.js
```diff
@@ -33,37 +35,31 @@ export default function Tasks(props: Props): React.Node {
...
-                animated: true,
-            });
-            currentIndexRef.current = currentTaskIndex;
+        if (totalSwipedTasks + 1 > currentTaskIndex) {
+            setTimeout(() => {
+                flatListRef.current?.scrollToIndex({
+                    index: currentTaskIndex,
+                    animated: true,
+                });
+            }, 0);
```
tnagorra, line 55: Let's add a NOTE here why we had to add a timeout

tnagorra (changes requested):
Let's use a dictionary instead of array to handle loading states

tnagorra (approved):

